PR #2, opened 2016-12-07
Enable UART during initialization on Zynq7000 platform

agacek:
The Zynq7000 UART is not initialized by default. On real hardware U-boot
will perform the correct initialization, but in QEMU there is no
initialization done. This commit enables the UART in elfloader so that
it will work both on the hardware and in QEMU.

### DIFF
--- a/elfloader-tool/src/arch-arm/plat-zynq7000/sys_fputc.h
+++ b/elfloader-tool/src/arch-arm/plat-zynq7000/sys_fputc.h
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2014, NICTA
+ *
+ * This software may be distributed and modified according to the terms of
+ * the GNU General Public License version 2. Note that NO WARRANTY is provided.
+ * See "LICENSE_GPLv2.txt" for details.
+ *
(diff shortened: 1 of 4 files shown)


elfloader-tool/src/arch-arm/plat-zynq7000/sys_fputc.h
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2014, NICTA
```
AdrianDanis: The NICTA license header is no longer used. Currently we add new files with

```c
/*
 * Copyright 2016, Data61
 * Commonwealth Scientific and Industrial Research Organisation (CSIRO)
 * ABN 41 687 119 230.
 *
 * This software may be distributed and modified according to the terms of
 * the GNU General Public License version 2. Note that NO WARRANTY is provided.
 * See "LICENSE_GPLv2.txt" for details.
 *
 * @TAG(D61_GPL)
 */
```

If you would be so kind as to change that

agacek:
I changed the license on the file I added. The old one is still on some of the older files.

PR #5, opened 2017-06-20
Calling cc-option, etc, with the configured compiler

mokshasoft:
The Kbuild function cc-option, that is used to test if a certain compiler
flag is supported by the compiler, was called with the wrong compiler if
the variable CONFIG_KERNEL_COMPILER was set to anything else but gcc. The
CC variable used by cc-option was set using CONFIG_KERNEL_COMPILER after
all calls to cc-option, as-option, ld-option, etc had been done. So these
commands always executed with the default gcc compiler and not the configured
one.

Using the cc-option is only sensible when an option may not be supported
by all compilers, so assigning the result of _one_ cc-option call to a variable
that is used by both the kernel and user space is incorrect. The kernel and
user space can be compiled using different compilers by setting the variables
CONFIG_(KERNEL|USER)_COMPILER, so calling cc-option with one of these compilers
and then also assigning the result to the other compiler's options, may lead to
that compiler not understanding the option.

This is quite a big change in the build system, but in order to be able to use one compiler for the kernel and one for user space it seems necessary to do it.

Running the sel4test on these four configurations pass.

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat simulate.sh 
#!/bin/sh
make mrproper ; make x64_simulation_debug_xml_defconfig ; make
make simulate-x86_64
make mrproper ; make ia32_simulation_debug_xml_defconfig ; make
make simulate-ia32
make mrproper ; make kzm_simulation_release_xml_defconfig ; make
make simulate-kzm
make mrproper ; make zynq_simulation_release_xml_defconfig ; make
make simulate-zynq

If build where deterministic it would have been a lot easier to test this patch, but they are not unfortunately (https://github.com/seL4/sel4test/issues/8).

To test that I haven't broken anything I have run make compile-all with a patched compile-all.sh script (see repo diff below) that builds the whole sel4test and saves the logs. I did one compile-all without the patch and one with it. All targets passed compilation. If the build logs are the same for all defconfigs, excluding white-space differences, I assume that nothing is broken, at least not in the sel4test repo. In order to get all the logs (also from the recursive make calls) from the build I had to intercept all invocations of make and add --debug=v V=2 (see make script below that calls a copy of itself adding the parameters). WARNING: Other gits that use cc-option, etc, but that is not included in sel4test will be broken if using the cc-option, and need to add the compiler parameter.

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat /usr/bin/make
#!/bin/sh
make-my-copy --debug=v V=2 $@

All the build logs for all defconfigs where the same except for the ia32 and x64 ones. Examining the differences in the build logs for ia32 and x64 shows that it is only the order of the compiler switches that differ. To prove that (at least to myself) I wrote a script that sorted each line and compared the result. With that definition of equality also the ia32 and x64 build logs matched. See the output of the compare-diffs.sh script just below.

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat compare-diffs.sh 
#!/bin/sh
compare_files () {
    left=$1
    right=$2

    # extract the difference between two logs
    diff -w $left $right > diff.txt

    # compare the differences, and check that each line is equal when sorted
    diff \
        <(grep "^<" diff.txt | sed 's/^..//g' |
          while read line; do echo $line | sed 's/ /\n/g' | sort | awk '{line=line " " $0} END {print line}' ; done) \
        <(grep "^>" diff.txt | sed 's/^..//g' |
          while read line; do echo $line | sed 's/ /\n/g' | sort | awk '{line=line " " $0} END {print line}' ; done)
}

folder1=$1
folder2=$2

# find all defconfigs where the build log is not exactly the same except for white-space differences
defconfigs=$(diff -qwr $folder1 $folder2 | awk '{print $2'} | sed 's/.*\///g')

for i in $defconfigs ; do
    echo "comparing logs for defconfigs $i"
    compare_files $folder1/$i $folder2/$i
done
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ . compare-diffs.sh buildlogs/220703/ buildlogs/232551/
comparing logs for defconfigs ia32_debug_xml_defconfig.log
comparing logs for defconfigs ia32_release_xml_defconfig.log
comparing logs for defconfigs ia32_release_xml_pae_defconfig.log
comparing logs for defconfigs ia32_simulation_debug_xml_defconfig.log
comparing logs for defconfigs ia32_simulation_release_xml_defconfig.log
comparing logs for defconfigs ia32_simulation_verification_xml_defconfig.log
comparing logs for defconfigs ia32_smp_debug_xml_defconfig.log
comparing logs for defconfigs ia32_smp_release_xml_defconfig.log
comparing logs for defconfigs ia32_smp_verification_xml_defconfig.log
comparing logs for defconfigs ia32_verification_xml_defconfig.log
comparing logs for defconfigs ia32_verification_xml_pae_defconfig.log
comparing logs for defconfigs x64_debug_xml_defconfig.log
comparing logs for defconfigs x64_release_xml_defconfig.log
comparing logs for defconfigs x64_simulation_debug_xml_defconfig.log
comparing logs for defconfigs x64_simulation_release_xml_defconfig.log
comparing logs for defconfigs x64_simulation_verification_xml_defconfig.log
comparing logs for defconfigs x64_smp_debug_xml_defconfig.log
comparing logs for defconfigs x64_smp_release_xml_defconfig.log
comparing logs for defconfigs x64_smp_verification_xml_defconfig.log
comparing logs for defconfigs x64_verification_xml_defconfig.log

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ repo diff
project projects/sel4test/
diff --git a/tools/compile-all.sh b/tools/compile-all.sh
index ba7d4ca..1113efe 100755
--- a/tools/compile-all.sh
+++ b/tools/compile-all.sh
@@ -12,13 +12,16 @@
 
 exit_code=0
 
+folder=buildlogs/$(date +%H%M%S)
+mkdir -p $folder
+
 for config in `/bin/ls -H -I "bamboo*" configs`
 do
     echo $config
     make clean > /dev/null
     make $config > /dev/null
     make silentoldconfig > /dev/null
-    env -i PATH="$PATH" make kernel_elf > /dev/null
+    env -i PATH="$PATH" make > $folder/$config.log
     if [ $? -eq 0 ]
     then
         echo "pass"

I also wanted to make sure that I haven't forgotten to edit the cc-option in any places, so I ran a script to try to find such places. The search shows mostly comments, and in the other lines I haven't found any forgotten places.

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat find-forgotten-cc-options.sh 
#!/bin/sh
# exclude build and git directories
EXCL="--exclude-dir=build --exclude-dir=stage --exclude-dir=.git"
# check that all calls to the option functions have gotten the compiler added
grep -r cc-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r cc-ldoption * ${EXCL} | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r ld-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r as-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r as-instr * ${EXCL}    | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
# find lines with more than one option call
grep -r "\$(call\ .*\$(call\ " * ${EXCL}
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ . find-forgotten-cc-options.sh 
common-tool/project.mk:# Kernel cc-option compiler
common-tool/project.mk:# Non-kernel cc-option compiler
kbuild-tool/Kbuild.include:# cc-option
kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call cc-option, <compiler>, -march=winchip-c6, -march=i586)
kbuild-tool/Kbuild.include:cc-option = $(call try-run,\
kbuild-tool/Kbuild.include:# cc-option-yn
kbuild-tool/Kbuild.include:# Usage: flag := $(call cc-option-yn, <compiler>, -march=winchip-c6)
kbuild-tool/Kbuild.include:cc-option-yn = $(call try-run,\
kbuild-tool/Kbuild.include:# cc-option-align
kbuild-tool/Kbuild.include:cc-option-align = $(subst -functions=0,,\
kbuild-tool/Kbuild.include:	$(call cc-option,$(1),-falign-functions=0,-malign-functions=0))
projects/tools/common-tool/project.mk:# Kernel cc-option compiler
projects/tools/common-tool/project.mk:# Non-kernel cc-option compiler
projects/tools/kbuild-tool/Kbuild.include:# cc-option
projects/tools/kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call cc-option, <compiler>, -march=winchip-c6, -march=i586)
projects/tools/kbuild-tool/Kbuild.include:cc-option = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# cc-option-yn
projects/tools/kbuild-tool/Kbuild.include:# Usage: flag := $(call cc-option-yn, <compiler>, -march=winchip-c6)
projects/tools/kbuild-tool/Kbuild.include:cc-option-yn = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# cc-option-align
projects/tools/kbuild-tool/Kbuild.include:cc-option-align = $(subst -functions=0,,\
projects/tools/kbuild-tool/Kbuild.include:	$(call cc-option,$(1),-falign-functions=0,-malign-functions=0))
kbuild-tool/Kbuild.include:# cc-ldoption
kbuild-tool/Kbuild.include:# Usage: ldflags += $(call cc-ldoption, <compiler>, -Wl$(comma)--hash-style=both)
kbuild-tool/Kbuild.include:cc-ldoption = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# cc-ldoption
projects/tools/kbuild-tool/Kbuild.include:# Usage: ldflags += $(call cc-ldoption, <compiler>, -Wl$(comma)--hash-style=both)
projects/tools/kbuild-tool/Kbuild.include:cc-ldoption = $(call try-run,\
kbuild-tool/Kbuild.include:# ld-option
kbuild-tool/Kbuild.include:# Usage: LDFLAGS += $(call ld-option, <compiler>, -X)
kbuild-tool/Kbuild.include:ld-option = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# ld-option
projects/tools/kbuild-tool/Kbuild.include:# Usage: LDFLAGS += $(call ld-option, <compiler>, -X)
projects/tools/kbuild-tool/Kbuild.include:ld-option = $(call try-run,\
kbuild-tool/Kbuild.include:# as-option
kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call as-option, <compiler>, -Wa$(comma)-isa=foo,)
kbuild-tool/Kbuild.include:as-option = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# as-option
projects/tools/kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call as-option, <compiler>, -Wa$(comma)-isa=foo,)
projects/tools/kbuild-tool/Kbuild.include:as-option = $(call try-run,\
kbuild-tool/Kbuild.include:# as-instr
kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call as-instr, <compiler>, instr, option1, option2)
kbuild-tool/Kbuild.include:as-instr = $(call try-run,\
projects/tools/kbuild-tool/Kbuild.include:# as-instr
projects/tools/kbuild-tool/Kbuild.include:# Usage: cflags-y += $(call as-instr, <compiler>, instr, option1, option2)
projects/tools/kbuild-tool/Kbuild.include:as-instr = $(call try-run,\
common-tool/Makefile.flags:	NK_CCFLAGS += $(call cc-option,$(NK_CC),-g3,-g) $(call cc-option,$(NK_CC),-ggdb3,-ggdb) $(call cc-option,$(NK_CC),-save-temps,)
projects/tools/common-tool/Makefile.flags:	NK_CCFLAGS += $(call cc-option,$(NK_CC),-g3,-g) $(call cc-option,$(NK_CC),-ggdb3,-ggdb) $(call cc-option,$(NK_CC),-save-temps,)
projects/tools/kbuild-tool/Makefile.lib:__c_flags	= $(call addtree,-I$(obj)) $(call flags,_c_flags)

### DIFF
--- a/common-tool/Makefile.flags
+++ b/common-tool/Makefile.flags
@@ -26,22 +26,27 @@ NK_ASFLAGS =
 
 NK_CCFLAGS =
 
-ldflags-y += $(call cc-ldoption, -Wl$(comma)--build-id=none)
+LDFLAGS += $(call cc-ldoption, $(K_CC), -Wl$(comma)--build-id=none)
+NK_LDFLAGS += $(call cc-ldoption, $(NK_CC), -Wl$(comma)--build-id=none)
 
 ifeq ($(CONFIG_ARCH_X86),y)
-	ccflags-y += $(call cc-option,-fno-stack-protector)
+	CCFLAGS += $(call cc-option, $(K_CC), -fno-stack-protector)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -fno-stack-protector)
 	export ARCH = x86
 	DEFINES += ARCH_X86
 	export ARCH_X86
 	export RUST_ENDIAN = little
 	export RUST_ENV = gnu
-	ccflags-y += $(call cc-option,-fno-pic)
+	CCFLAGS += $(call cc-option, $(K_CC), -fno-pic)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -fno-pic)
 	ldflags-y += -static
 endif
 
 ifeq ($(CONFIG_ARCH_ARM),y)
-	ccflags-y += $(call cc-option,-marm)
-	ccflags-y += $(call cc-option,-Wtrampolines)
+	CCFLAGS += $(call cc-option, $(K_CC), -marm)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -marm)
+	CCFLAGS += $(call cc-option, $(K_CC), -Wtrampolines)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -Wtrampolines)
 	export ARCH = arm
 	DEFINES += ARCH_ARM
 	export ARCH_ARM
@@ -50,7 +55,8 @@ ifeq ($(CONFIG_ARCH_ARM),y)
 endif
 
 ifeq ($(CONFIG_ARCH_IA32),y)
-	ccflags-y += $(call cc-option,-m32)
+	CCFLAGS += $(call cc-option, $(K_CC), -m32)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -m32)
 
 	ldflags-y += -Wl,-m,elf_i386
 	asflags-y += -Wa,--32
@@ -71,8 +77,10 @@ ifeq ($(CONFIG_ARCH_IA32),y)
 endif
 
 ifeq ($(CONFIG_ARCH_X86_64),y)
-	cflags-y += $(call cc-option,-m64)
-	cflags-y += $(call cc-option, -fno-asynchronous-unwind-tables,)
+	CFLAGS += $(call cc-option, $(K_CC), -m64)
+	NK_CFLAGS += $(call cc-option, $(NK_CC), -m64)
+	CFLAGS += $(call cc-option, $(K_CC), -fno-asynchronous-unwind-tables,)
+	NK_CFLAGS += $(call cc-option, $(NK_CC), -fno-asynchronous-unwind-tables,)
 	ldflags-y  += -Wl,-m -Wl,elf_x86_64
 	asflags-y  += -Wa,--64
 	NK_LDFLAGS += -Wl,-m -Wl,elf_x86_64
@@ -147,8 +155,10 @@ ifeq ($(CONFIG_ARCH_AARCH64),y)
 endif
 
 ifeq ($(CONFIG_ARM1136JF_S),y)
-	ccflags-y += $(call cc-option,-mcpu=arm1136jf-s,-march=armv6)
-	asflags-y += $(call as-option, -Wa$(comma)-mcpu=arm1136jf-s, -Wa$(comma)-march=armv6)
+	CCFLAGS += $(call cc-option, $(K_CC), -mcpu=arm1136jf-s, -march=armv6)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -mcpu=arm1136jf-s, -march=armv6)
+	ASFLAGS += $(call as-option, $(K_CC), -Wa$(comma)-mcpu=arm1136jf-s, -Wa$(comma)-march=armv6)
+	NK_ASFLAGS += $(call as-option, $(NK_CC), -Wa$(comma)-mcpu=arm1136jf-s, -Wa$(comma)-march=armv6)
 	DEFINES += ARMV6
 	DEFINES += ARM1136J_S
 	export RUST_CPU = arm1136jf-s
@@ -159,8 +169,10 @@ endif
 
 
 ifeq ($(CONFIG_ARM_CORTEX_A7),y)
-	ccflags-y += $(call cc-option, -mcpu=cortex-a7,-march=armv7-a)
-	asflags-y += $(call as-option, -Wa$(comma)-mcpu=cortex-a7, -Wa$(comma)-march=armv7-a)
+	CCFLAGS += $(call cc-option, $(K_CC), -mcpu=cortex-a7, -march=armv7-a)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -mcpu=cortex-a7, -march=armv7-a)
+	ASFLAGS += $(call as-option, $(K_CC), -Wa$(comma)-mcpu=cortex-a7, -Wa$(comma)-march=armv7-a)
+	NK_ASFLAGS += $(call as-option, $(NK_CC), -Wa$(comma)-mcpu=cortex-a7, -Wa$(comma)-march=armv7-a)
 	DEFINES += ARMV7_A
 	DEFINES += ARM_CORTEX_A7
 	export ARMV=armv7-a
@@ -168,8 +180,10 @@ ifeq ($(CONFIG_ARM_CORTEX_A7),y)
 endif
 
 ifeq ($(CONFIG_ARM_CORTEX_A8),y)
-	ccflags-y += $(call cc-option, -mcpu=cortex-a8,-march=armv7-a)
-	asflags-y += $(call as-option, -Wa$(comma)-mcpu=cortex-a8, -Wa$(comma)-march=armv7-a)
+	CCFLAGS += $(call cc-option, $(K_CC), -mcpu=cortex-a8,-march=armv7-a)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -mcpu=cortex-a8,-march=armv7-a)
+	ASFLAGS += $(call as-option, $(K_CC), -Wa$(comma)-mcpu=cortex-a8, -Wa$(comma)-march=armv7-a)
+	NK_ASFLAGS += $(call as-option, $(NK_CC), -Wa$(comma)-mcpu=cortex-a8, -Wa$(comma)-march=armv7-a)
 	DEFINES += ARMV7_A
 	DEFINES += ARM_CORTEX_A8
 	export ARMV=armv7-a
@@ -179,8 +193,10 @@ ifeq ($(CONFIG_ARM_CORTEX_A8),y)
 endif
 
 ifeq ($(CONFIG_ARM_CORTEX_A9),y)
-	ccflags-y += $(call cc-option, -mcpu=cortex-a9, -march=armv7-a)
-	asflags-y += $(call as-option, -Wa$(comma)-mcpu=cortex-a9, -Wa$(comma)-march=armv7-a)
+	CCFLAGS += $(call cc-option, $(K_CC), -mcpu=cortex-a9, -march=armv7-a)
+	NK_CCFLAGS += $(call cc-option, $(NK_CC), -mcpu=cortex-a9, -march=armv7-a)
+	ASFLAGS += $(call as-option, $(K_CC), -Wa$(comma)-mcpu=cortex-a9, -Wa$(comma)-march=armv7-a)
+	NK_ASFLAGS += $(call as-option, $(NK_CC), -Wa$(comma)-mcpu=cortex-a9, -Wa$(comma)-march=armv7-a)
 	DEFINES += ARMV7_A
 	DEFINES += ARM_CORTEX_A9
 	export ARMV=armv7-a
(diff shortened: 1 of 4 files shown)


mokshasoft:
Oh, I can see that my scripts were not interpreted correctly in the comment above. Second try...

```
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ repo diff
project projects/sel4test/
diff --git a/tools/compile-all.sh b/tools/compile-all.sh
index ba7d4ca..1113efe 100755
--- a/tools/compile-all.sh
+++ b/tools/compile-all.sh
@@ -12,13 +12,16 @@
 
 exit_code=0
 
+folder=buildlogs/$(date +%H%M%S)
+mkdir -p $folder
+
 for config in `/bin/ls -H -I "bamboo*" configs`
 do
     echo $config
     make clean > /dev/null
     make $config > /dev/null
     make silentoldconfig > /dev/null
-    env -i PATH="$PATH" make kernel_elf > /dev/null
+    env -i PATH="$PATH" make > $folder/$config.log
     if [ $? -eq 0 ]
     then
         echo "pass"
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat simulate.sh 
#!/bin/sh
make mrproper ; make x64_simulation_debug_xml_defconfig ; make
make simulate-x86_64
make mrproper ; make ia32_simulation_debug_xml_defconfig ; make
make simulate-ia32
make mrproper ; make kzm_simulation_release_xml_defconfig ; make
make simulate-kzm
make mrproper ; make zynq_simulation_release_xml_defconfig ; make
make simulate-zynq

sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat compare-diffs.sh 
#!/bin/sh

compare_files () {
    left=$1
    right=$2

    # extract the difference between two logs
    diff -w $left $right > diff.txt

    # compare the differences, and check that each line is equal when sorted
    diff \
        <(grep "^<" diff.txt | sed 's/^..//g' |
          while read line; do echo $line | sed 's/ /\n/g' | sort | awk '{line=line " " $0} END {print line}' ; done) \
        <(grep "^>" diff.txt | sed 's/^..//g' |
          while read line; do echo $line | sed 's/ /\n/g' | sort | awk '{line=line " " $0} END {print line}' ; done)
}

folder1=$1
folder2=$2

# find all defconfigs where the build log is not exactly the same except for white-space differences
defconfigs=$(diff -qwr $folder1 $folder2 | awk '{print $2'} | sed 's/.*\///g')

for i in $defconfigs ; do
    echo "comparing logs for defconfigs $i"
    compare_files $folder1/$i $folder2/$i
done
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat /usr/bin/make
#!/bin/sh
make-my-copy --debug=v V=2 $@
sel4@sel4:~/Repo/seL4/TestSuite/sel4test3$ cat find-forgotten-cc-options.sh 
#!/bin/sh
# exclude build and git directories
EXCL="--exclude-dir=build --exclude-dir=stage --exclude-dir=.git"
# check that all calls to the option functions have gotten the compiler added
grep -r cc-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r cc-ldoption * ${EXCL} | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r ld-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r as-option * ${EXCL}   | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
grep -r as-instr * ${EXCL}    | grep -v "\$(NK_CC)" | grep -v "\$(K_CC)"
# find lines with more than one option call
grep -r "\$(call\ .*\$(call\ " * ${EXCL}
```

mokshasoft:
This patch (or another that solves the same problem) is necessary to compile the kernel with clang. When setting the CONFIG_KERNEL_COMPILER to "clang" still the cc-options are called with gcc. This later creates build problems when the gcc flags are passed to clang. I guess that this is a quite big patch, so your comments on it or suggestions of other solutions would be great.

AdrianDanis:
I don't disagree that the current option checking is incorrect, but an alternate solution would be to remove support for compiling the kernel with a different compiler, and now all the option setting becomes correct. That is not meant to be a complete joke suggestion, as internally we have been slowly developing a new build system and it does not presently have support for compiling the kernel with a different compiler.

I think the first question that should be asked is why are you wanting to compile only the kernel with clang? Is there a reason not to compile user space with clang as well?

mokshasoft:
My first reason to go for compiling only the kernel with clang was because I thought that would be a smaller task. I like doing things in smaller steps generally, and after that was done I was going to go for user space. But maybe now I understand that it was not smaller... :)

The second reason was because the two options actually was there, I thought that it had been put there for a reason.

After digging in the makefiles I also got the thought to rewrite the build system, but that was a future task that I was thinking about. It sounds good that you actually have started doing that! Compile time is very slow now compared what it could be. Which build system have you chosen? I thought about Shake, it seems to be good, but the reason was also because I wanted to try a real problem on it once.

AdrianDanis:
It is definitely fair to want to use an option that is there. Somewhat our fault for breaking support for it over the years and not removing it. Not that I am particularly against this change to fix it, although the duplication for setting flags is a bit yuck. Could be solved by yet more makefile macros, but I'm not sure they are ever really the right solution.

Do you have any guesses as to how hard it would be to get the rest of user space compiling with clang? If you're planning to update user space anyway, and eventually make this change unnecessary, then my preference would be to not try and resuscitate this feature.  

CMake is the basis of the in development build system. Although I cannot say when it will appear, it will be a completely parallel build system when it does.

mokshasoft:
CMake is also a good option in my opinion. Is there any chance that this internal CMake branch will go public so that people can help in the porting work? At the moment I am in the middle of a move, and a quite scheduled summer, so I don't know how much I would be able to help, but it would definitely be something that I would find interesting.

I guess that compiling user space with clang is fairly straight forward now when I basically have a prototype for compiling most kernel targets with clang. User space should be very similar for most targets, so I guess that it is not so difficult. I'll definitely go along that path then.

AdrianDanis:
It will definitely be made public before everything is ported to it, just need to be sure it is a good foundation first.

PR #6, opened 2017-07-05
Secure Load & IBLL

chrisguikema:
Secure Load is a way to ensure that what was built by the developer is not tampered with before loading onto the target platform.

In the original build process, the Kernel and the Application are stored into an archive for the elfloader to unpack. This process performs an md5 or a sha256 sum on both of the files and stores the resulting checksum in the archive with the images. The elfloader was modified to perform the same sum on the images, and compare the calculated sum and the sum in the archive. The sum is also printed out during the build process and during loading, so even if someone changed the executable's bits and the sum image in the archive, the difference could be noticed by the developer by a simple eye test.

Because the elfloader Kconfig isn't accessable in most of the top-level Kconfigs, there are two ways to invoke the hashing algorithms:
      1. CONFIG_HASH_INSTRUCTIONS=y in your defconfig
         CONFIG_HASH_SHA (optional - remove if you want to md5 hash)
      2. Include the elf-loader's Kconfig under a new menu in your top-level Kconfig
         like this:
         menu "Tools"
           source "tools/elfloader/Kconfig"
         endmenu

The makefile was modified to ensure the proper variables were exported when the gen_boot_image.sh script is run. 

A pre-image command was also added which allows the user to link a pre-built application image for an seL4 application (not CAmkES). This allows developers to collaborate without sharing source code and intellectual property. 

This entire process is called Independent Build, Link, and Load (IBLL), since the Kernel, Libraries, and Applications are all built separately and can be built at different times and by different developers, then linked and reliably loaded onto the targets.

### DIFF
--- a/common-tool/project-arm.mk
+++ b/common-tool/project-arm.mk
@@ -39,6 +39,7 @@ elfloader: libcpio common FORCE
 %-image: export TOOLPREFIX=$(CONFIG_CROSS_COMPILER_PREFIX:"%"=%)
 %-image: export STRIP=$(CONFIG_REMOVE_SYMBOLS)
 %-image: export HASH=$(CONFIG_HASH_INSTRUCTIONS)
+%-image: export HASH_SHA=$(CONFIG_HASH_SHA)
 %-image: export V
 %-image: % kernel_elf common elfloader FORCE
 	@echo "[GEN_IMAGE] $@-$(ARCH)-$(PLAT)"
--- a/common-tool/project-arm.mk
+++ b/common-tool/project-arm.mk
@@ -37,11 +37,26 @@ elfloader: libcpio common FORCE
 	@echo "[elfloader] done."
 
 %-image: export TOOLPREFIX=$(CONFIG_CROSS_COMPILER_PREFIX:"%"=%)
+%-image: export STRIP=$(CONFIG_REMOVE_SYMBOLS)
+%-image: export HASH=$(CONFIG_HASH_INSTRUCTIONS)
 %-image: export V
 %-image: % kernel_elf common elfloader FORCE
 	@echo "[GEN_IMAGE] $@-$(ARCH)-$(PLAT)"
 	$(Q)$(SEL4_COMMON)/elfloader/gen_boot_image.sh $(STAGE_BASE)/kernel.elf \
 		$(STAGE_BASE)/bin/$< $(IMAGE_ROOT)/$@-$(ARCH)-$(PLAT) 2>&1 \
+		| while read line; do echo " [GEN_IMAGE] $$line"; done; \
+		exit $${PIPESTATUS[0]}
+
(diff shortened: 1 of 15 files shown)


elfloader-tool/src/arch-arm/main.c
```diff
@@ -148,32 +204,8 @@ static paddr_t load_elf(const char *name, void *elf,
     info->virt_entry = (vaddr_t)elf_getEntryPoint(elf);
     info->phys_virt_offset = dest_paddr - (vaddr_t)min_vaddr;
 
-    /* Round up the destination address to the next page */
```
yyshen: Why is this removed?

elfloader-tool/src/arch-arm/utils/hash.c
```diff
@@ -0,0 +1,52 @@
...
+        md5_update(&calculated_hash, file_to_hash, bytes_to_hash);
+        md5_sum(&calculated_hash, outputted_hash);
+    }
+}
+
+/* Function to print the hash */
+void print_hash(void *hash_to_print, int bytes_to_print)
+{
+    for(int i = 0; i < bytes_to_print; i++) {
+        printf("%02x",*(uint8_t *)hash_to_print++);
```
yyshen: Correct me if I am wrong. C standard does not allow arithmetic on void * but GCC allows it as an extension. Maybe we should rephrase the printf.

elfloader-tool/src/arch-arm/main.c
chrisguikema: We started this work a while ago on an older commit. It may have been removed when rebasing the IBLL and Secure Load. Putting it back.

elfloader-tool/src/arch-arm/utils/hash.c
chrisguikema: Changed to use a uint8_t pointer as the input

chrisguikema:
I went through and fixed the initial comments with a rebase

pingerino:
Nice! I've tested this and merged it internally with a few changes (readded -Wall in the Makefile). You should see it come out in the next day or two as it goes through our tests and gets pushed out to github.

common-tool/project-arm.mk
```diff
@@ -45,6 +47,19 @@ elfloader: libcpio common FORCE
 		| while read line; do echo " [GEN_IMAGE] $$line"; done; \
 		exit $${PIPESTATUS[0]}
 
+# make pre-image IMAGE_DIR=~/sel4-tutorials/stage/arm/zynq7000/bin/hello-3
```
kent-mcleod: Are we able to delete this pre-image rule?  It appears to be left over from testing.

Furao: Yes, that can be removed. Would you like us to update the pull request?

pingerino: I've updated the PR internally. You should see it come through tonight.

pingerino:
The changes are merged, however github hasn't picked it up as I rebased them.

PR #7, opened 2017-09-14
Zcu102 7.0.0

Furao:
This pull request adds AARCH32 and AARCH64 support for the Xilinx Zynq UltraScale+ MPSoC. Tested on the ZCU102 development board.

### DIFF
--- a/common-tool/Makefile.flags
+++ b/common-tool/Makefile.flags
@@ -306,6 +306,11 @@ ifeq ($(CONFIG_PLAT_ZYNQ7000),y)
 	export PLAT=zynq7000
 endif
 
+ifeq ($(CONFIG_PLAT_ZYNQMP),y)
+	DEFINES += PLAT_ZYNQMP
+	export PLAT=zynqmp
+endif
+
 ifeq ($(CONFIG_PLAT_HIKEY),y)
 	DEFINES += HIKEY
 	export PLAT=hikey
@@ -347,7 +352,7 @@ endif
 ifneq (,$(filter $(PLAT), tk1 tx1))
 export MACH=nvidia
 else
-export MACH=$(strip $(foreach mach,exynos imx omap,$(findstring $(mach),$(PLAT))))
+export MACH=$(strip $(foreach mach,exynos imx omap zynq,$(findstring $(mach),$(PLAT))))
 endif
 
 # Kernel debugging.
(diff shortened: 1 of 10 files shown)


Furao:
Feel free to drop the "Added DARPA License" commit. We just need to retain it in our own repositories.

elfloader-tool/src/arch-arm/armv/armv8-a/64/mmu-hyp.S
```diff
@@ -51,6 +51,10 @@ BEGIN_FUNC(leave_hyp)
     mov     x19, #(PSR_F_BIT | PSR_I_BIT | PSR_A_BIT | PSR_D_BIT | PSR_MODE_EL1h)
     msr     spsr_el2, x19
 
+    /* Enable FP for EL1 */
+    mov     x19, (3 << 20)
+    msr     cpacr_el1, x19
```
AdrianDanis: The commit message attached to this change suggests you are enabling the use of FPU registers in the kernel. This is not correct to do as our FPU management code assumes that the kernel does not use the FPU registers and we pass flags when building the kernel to prevent the compiler from using such registers.

natestuder: The elfloader doesn't seem to use the same flags, so it can end up using the FPU which causes exceptions.

If we use u-boot and enable hypervisor support, this issue gets hidden since u-boot enables the FPU for EL2.  If we try running without either of those it doesn't work.

Would it be better to add similar flags to the elfloader?

AdrianDanis: I think it might be best to just add the [same flags](https://github.com/seL4/seL4/blob/master/Makefile#L467) to the elfloader, as there's no real reason for the elfloader to be using the FPU and I think it's simpler going forward if it doesn't.

Furao: I will push update the request with the suggested changes.

mikedefrancis: I have seen exceptions from the elfloader on the zynq9000 when these two asm lines are omitted. I will retest @Furao 's code when the flags are passed properly to the elfloader and see if my exception persists.

Furao:
Updated based on comments.

PR #15, opened 2018-07-22
RISCV Level 2 Mappings

chrisguikema:
These commits support building 64-bit seL4 for the Rocket-Chip (https://github.com/ucb-bar/fpga-images-zedboard/tree/7e0080ee9e29e399d80d74d9756f20cece7aeef0)

Since the Rocket-Chip only has 256 MiB of RAM, its physical address is no longer aligned on a 1GiB boundary. These commits clean up the existing elfloader code to make things more readable, fix a simple bug, and then add another array for level 2 mappings which gets filled in if the ROCKET_CHIP config option is is selected (that is defined in the kernel, which is another pull request)

### DIFF
--- a/elfloader-tool/src/arch-riscv/boot.c
+++ b/elfloader-tool/src/arch-riscv/boot.c
@@ -58,10 +58,14 @@ typedef uint64_t seL4_Word;
 
 #define GET_PT_INDEX(addr, n) (((addr) >> (((PT_INDEX_BITS) * ((CONFIG_PT_LEVELS) - (n))) + RISCV_PGSHIFT)) % PTES_PER_PT)
 
+#define VIRT_PHYS_ALIGNED(virt, phys, level_bits) (IS_ALIGNED((virt), (level_bits)) && IS_ALIGNED((phys), (level_bits)))
+
 struct image_info kernel_info;
 struct image_info user_info;
 
 unsigned long l1pt[PTES_PER_PT] __attribute__((aligned(4096)));
+unsigned long l2pt[PTES_PER_PT] __attribute__((aligned(4096)));
+
 char elfloader_stack_alloc[BIT(CONFIG_KERNEL_STACK_BITS)];
 
 void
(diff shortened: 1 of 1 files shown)


chrisguikema:
We're going to update this request based on Hesham's feedback from the kernel pull request.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -28,14 +35,28 @@
 // page table entry (PTE) field
 #define PTE_V     0x001 // Valid
 
-#define PTE64_PPN0_SHIFT 10
+#define PTE_PPN0_SHIFT 10
 
-#define PTES_PER_PT (RISCV_PGSIZE/sizeof(long))
+#if __riscv_xlen == 32
```
pingerino: It would be *nice* if these constants came from a include/arch-riscv/<32/64>/mode header, as we do for arm. I'm happy to fix this internally though after. 

For context we have been moving away from #ifdef based definitions, as we've found file includes to be less error prone and more reliable.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -28,14 +35,28 @@
...
+#define PPN_HIGH       20
+typedef uint32_t seL4_Word;
+#else
+#define PTE_LEVEL_BITS PT_LEVEL_1_BITS
+#define PT_INDEX_BITS  9
+#define PPN_HIGH       28
+typedef uint64_t seL4_Word;
+#endif
+
+#define PTES_PER_PT (1 << PT_INDEX_BITS)
```
pingerino: nitpick: please use the BIT macro here, as it

elfloader-tool/src/arch-riscv/boot.c
chrisguikema: @pingerino that is really good to know for future work. Feel free to make that change afterwards, and I'll make a note of it for future work.

elfloader-tool/src/arch-riscv/boot.c
chrisguikema: I'll update to use the BIT macro.

chrisguikema:
@pingerino For small changes like the BIT macro, could D61 make those changes? For anything larger, I can handle it.

pingerino:
@chrisguikema sure no problem. I'm trying to get these PRs through this week.

chrisguikema:
Awesome! Good to hear. Let me know if you have any questions.

pingerino:
this has also been merged with edited history (some squashes) so github can't tell it has happened

PR #16, opened 2018-07-26
Switch to use custom rust compiler to enable std rust, while keeping …

podhrmic:
…robigalia+xargo bare-metal targets


podhrmic:
This is all that is needed to enable Xargo+std builds

kent-mcleod:
Are you able to rebase out the compiler commit and its revert commit pair?

kent-mcleod:
I'm also happy to do this myself as I merge it if you want.

podhrmic:
I won't get to this until tomorrow, so it is up to you.

PR #19, opened 2018-11-15
Add clang build for x86-64

mokshasoft:
Add clang compilation for seL4 x86-64. This pull request is needed by, https://github.com/seL4/seL4/pull/105. That pull request also describes the tests that have been done.


axel-h:
Please add the script to the patch.

pingerino:
Hi @mokshasoft, thanks for your contribution. We're working on this internally and will use it as a starting point, but the change needs more work and testing on all of our architectures before we merge it.

mokshasoft:
Outdated patch, clang is probably already supported now.

PR #20, opened 2018-11-19
Imx6 smc

natestuder:
Fixes to some small issues found when using the monitor mode hook on the iMX6.

### DIFF
--- a/elfloader-tool/src/plat/imx6/monitor.S
+++ b/elfloader-tool/src/plat/imx6/monitor.S
@@ -65,10 +65,10 @@ smc_handler:
     push {lr}
     blx r0
     isb
-    mrc p15, 0, r5, c1, c1, 0
+    mrc p15, 0, r12, c1, c1, 0
     /* set the NS bit */
-    orr r5, r5, #1
-    mcr p15, 0, r5, c1, c1, 0
+    orr r12, r12, #1
+    mcr p15, 0, r12, c1, c1, 0
     pop {lr}
     isb
     movs pc, lr
--- a/elfloader-tool/src/plat/tk1/monitor.S
+++ b/elfloader-tool/src/plat/tk1/monitor.S
@@ -64,10 +64,10 @@ smc_handler:
     ldr sp, [pc, #-12]
     push {lr}
     blx  r0
-    mrc p15, 0, r0, c1, c1, 0
+    mrc p15, 0, r12, c1, c1, 0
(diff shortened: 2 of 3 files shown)


elfloader-tool/src/plat/tk1/monitor.S
```diff
@@ -64,10 +64,10 @@ smc_handler:
     ldr sp, [pc, #-12]
     push {lr}
     blx  r0
-    mrc p15, 0, r0, c1, c1, 0
+    mrc p15, 0, r3, c1, c1, 0
```
yyshen: If we stick to the calling convention, r3 is also a return result register. Maybe we could use r12 instead?

elfloader-tool/src/plat/imx6/monitor.S
```diff
@@ -65,10 +65,10 @@ smc_handler:
     push {lr}
     blx r0
     isb
-    mrc p15, 0, r5, c1, c1, 0
+    mrc p15, 0, r3, c1, c1, 0
     /* set the NS bit */
-    orr r5, r5, #1
-    mcr p15, 0, r5, c1, c1, 0
+    orr r3, r3, #1
```
yyshen: same. could we use r12 instead of r3?

yyshen:
Hi Nathan,

Thanks for the PR. Sorry for the arbitrary choice of registers and any painful debugging it may cause. The intention of the hook is to execute some a short sequence of instructions provided by the caller, so the caller should know which register should be saved. However, sticking to the calling convention would be helpful. I made some comments inline. Please let me know what you think.

elfloader-tool/src/plat/tk1/monitor.S
natestuder: That should work.  I'll send an updated commit.

pingerino:
hey @natestuder, I'm happy to merge this, looks like it got lost during the holidays. Do you mind if I rebase it to avoid a merge commit? Otherwise, can you please rebase and merge?

natestuder:
Rebased and tested.

kent-mcleod:
Hey @natestuder, the original commits were accidentally merged, (https://github.com/seL4/seL4_tools/commit/f5fd7dab0d50e84cb8a14f661faeda9d39e74c33).  Are you happy with that or do you want us to revert the commits and apply the ones in this PR?

natestuder:
Works for me.

PR #21, opened 2019-01-17
add build flags for rpi3 (idk if this is correct...)

tv-deusen:


tv-deusen:
Apologies wasn't paying enough attention

PR #25, opened 2019-08-14
Zcu102 smp

chrisguikema:
This commit just adds support for enabling SMP mode on the zynqmp. There's also a commit that disables the yaml loader warning.


kent-mcleod:
Hey @chrisguikema, 
I've tested this internally and happy to merge.  Are you happy to apply the following?:
- I don't think your top commit is needed any more since c1d011517f8 uses yaml.safe_load instead.
- Can you remove the repo change-id part from your commit messages.  Maybe replace it with something like: `Add required logic for booting 4 cores on the zynqmp`

chrisguikema:
Yeah, looks like the first commit isn't needed. I'll update the commit messages to remove the change-id hash.

PR #28, opened 2019-10-17
Shoehorn: Fixes type annotations and member access for libarchive

Jesse-Millwood:
The wrong type annotation was being used with get_bytes.
The "name" member was being used with archive objects but the correct
member name is pathname.

### DIFF
--- a/cmake-tool/helpers/shoehorn.py
+++ b/cmake-tool/helpers/shoehorn.py
@@ -29,6 +29,7 @@ import re
 import sys
 
 import libarchive
+import libarchive.public
 
 import elf_sift
 import platform_sift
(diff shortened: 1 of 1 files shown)


branden-data61:
Hi Jesse,

Can you share with me the failure or diagnostic messages you got from the Python interpreter that prompted these changes?  I may be using a different version of libarchive (I have `python3-libarchive-c` 2.8-0.3 from Debian 10 installed).

Regards,
Branden

Jesse-Millwood:
@branden-data61 
I had installed `libarchive` from `pip`, which seems to be a different package. I must have had an old install of sel4-deps.

PR #30, opened 2019-11-02
elfloader: unify imx8 platform

axel-h:
Unify the code to provide a generic imx8 platform. Note that this change also requires a change in seL4/seL4.git to use the generic imx8 platform. I'll provide the link here once the pull request has been created


axel-h:
Base on the information provided by @kent-mcleod  I'm dropping this pull request, as sub platforms will be handled differently.

PR #35, opened 2019-12-04
Don't attempt to build hosted programs to test GCC

maybe-sybr:
By building a program which provides `main()`, we were assuming that
there is a libc available to compiler. This is not the case for the Red
Hat provided cross-compilers which are intended for use in freestanding
environments (the example given is kernel code). Given that the seL4
build system assumes a freestanding environment it is sensible to only
check that the compiler being used is capable of building such programs.

This change alters the code built when testing compiler options in
`environment_flags.cmake`, and also adds fixed options to force the
compiler to not link against the stdlib or include crt objects. By doing
so we avoid any requirement for a target libc on the build host.

PR #36, opened 2019-12-09
add_fpu_compilation_options: More accurate checks

kent-mcleod:
The FPU checks need to check that the compiler can compile programs with
the desired FPU options AND link against compiler runtime libraries with
the correct FPU ABI. If either of these conditions aren't met, then we
can't build programs correctly.

We use the standard check_c_source_runs CMake function which inherits
the compilation options that have been set already. Unfortunately it
doesn't inherit any changes to CMAKE_<LANG>_LINK_EXECUTABLE variables
which is what we use to set the compiler libraries and so we need to
specify those again manually. This both compiles and links our test
program.  In cross-compiling environments, the run stage will be
skipped; yet this is still sufficient to pick up ABI mismatches.

### DIFF
--- a/cmake-tool/helpers/environment_flags.cmake
+++ b/cmake-tool/helpers/environment_flags.cmake
@@ -84,6 +84,16 @@ macro(find_libgcc_files)
     endif()
 endmacro()
 
+
+# Call check_c_source_runs but set the cache variables
+# that cause the generated executable to not be run
+# on cross_compilation targets.
+macro(check_c_source_runs_cross_compile program var)
+    set(${var}_EXITCODE 0 CACHE INTERNAL "")
+    set(${var}_EXITCODE__TRYRUN_OUTPUT "" CACHE INTERNAL "")
+    check_c_source_runs(${program} ${var})
+endmacro()
+
 macro(add_fpu_compilation_options)
     # We want to check what we can set the -mfloat-abi to on arm and if that matches what is requested
     if(KernelSel4ArchAarch64)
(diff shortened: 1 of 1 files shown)


nomadeel:
@ssrg-bamboo `test`

ssrg-bamboo:
Hello, I'm a bot! I'll bring this PR into Trustworthy Systems and run some tests

kent-mcleod:
@maybe-sybr, the changes in https://github.com/seL4/seL4_tools/pull/35 cause the test programs to no longer get linked against libgcc or the crt files which test compatible floating point abis.  This PR should still work on systems without an installed cross- C library but I can only test it on our docker container configurations.

ssrg-bamboo:
All the tests we ran have passed! Nice job!

maybe-sybr:
lgtm, seems much more sensible to make cmake do the heavy lifting for us!

It looks like the adder camkes app continues to build and function fine for me targeting zynq7k in a clean tree with the Fedora xcompilers:

```
$ ../init-build.sh -DPLATFORM=zynq7000 -DCROSS_COMPILER_PREFIX=arm-linux-gnu-
...
-- Performing Test HARD_FLOAT
-- Performing Test HARD_FLOAT - Success
-- Performing Test compiler_arch_test
-- Performing Test compiler_arch_test - Success
-- Detecting cached version of: musllibc
--   Not found cache entry for musllibc - will build from source
...
$ ninja
...
$ ./simulate
...
Enabling MMU and paging
Bootstrapping kernel
Warning: Could not infer GIC interrupt target ID, assuming 0.
Booting all finished, dropped to user space
client: what's the answer to 342 + 74 + 283 + 37 + 534 ?
adder: Adding 342
adder: Adding 74
adder: Adding 283
adder: Adding 37
adder: Adding 534
client: result was 1270
QEMU: Terminated
```

PR #46, opened 2020-10-23
Replace legacy BBL with OpenSBI

alistair23:
This is part of a collection of PRs to convert from using the legacy BBL and riscv-pk to OpenSBI.

Other PRs include:

Signed-off-by: Alistair Francis <alistair.francis@wdc.com>


cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,28 @@ function(DeclareRootserver rootservername)
...
                     COMMAND
-                        cd ${CMAKE_BINARY_DIR}/bbl && ${BBL_PATH}/configure
-                        --quiet
-                        --host=${host}
-                        --with-arch=${march}
-                        --with-payload=${elf_target_file}
-                            && make -s clean && make -s > /dev/null
+                        cd ${OPENSBI_PATH} && ${CMAKE_OBJCOPY} -O binary ${elf_target_file}
+                        "payload" && make -s clean && CROSS_COMPILE=${CROSS_COMPILER_PREFIX} make
+                        PLATFORM=sifive/fu540 FW_PAYLOAD_PATH=payload
```
szhuang, line 115: It seems that this command is for Hifive Unleashed board only. Besides Hifive, we also need to support Ariane, qemu/spike and rocketchip(mapped to ZC706 and ZCU102).

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,28 @@ function(DeclareRootserver rootservername)
...
                     DEPENDS ${elf_target_file} elfloader ${USES_TERMINAL_DEBUG}
                 )
-                set(elf_target_file "${CMAKE_BINARY_DIR}/bbl/bbl")
+                set(
+                    elf_target_file
+                    "${OPENSBI_PATH}/build/platform/sifive/fu540/firmware/fw_payload.elf"
```
szhuang, line 120: See above

szhuang (changes requested):

cmake-tool/helpers/rootserver.cmake
alistair23: Hmm... Is the platform specified somewhere in the build scripts?

OpenSBI supports the Unleashed, Ariane and QEMU. So we could just change the platform. I don't see support for the rocket chip though.

szhuang: Variable ${PLATFORM} contains the name of the targeting platform. Regarding rocket chip, it merely needs HTIF from the bootloader. It seems that OpenSBI has some HTIF support, but I'm not sure how to use it.

Possible values of ${PLATFORM}: ariane, hifive, polarfire, spike, rocketchip.

Polarfire uses U540. I assume it can just use the command for the Unleashed.
Spike is QEMU simulation, it has to support both 64bit and 32bit. We use the "spike" machine for 64bit, but the "virt" machine for 32bit. I couldn't get the 32bit "spike" machine working. Do you have some insights?
```bash
# 32bit simulation, only works for virt machine
qemu-system-riscv32 -machine virt -cpu rv32 -nographic -serial mon:stdio -m size=2000M -bios none -kernel images/sel4test-driver-image-riscv-spike
```

alistair23: I just pushed an update that should handle the different platforms.

I don't have any insights sorry, the Spike machine is rarely used so using the virt machine is probably a better choice anyway.

alistair23:
Ping!

szhuang:
Looks good to me. I'll test it on hardware.

axel-h:
* please use the double quotes for the strings, especially if they contain variables the are paths that the user can set to whatever he/she has locally
* Can't we set the `OPENSBI_PLATFORM` string in every `<kernel>/src/plat/${you-name-it}/config.cmake`, so we kann keep `cmake-tool/helpers/rootserver.cmake` agnostic of the detail. It should just fail if `UseRiscVOpenSBI` is set, but `OPENSBI_PLATFORM` is missing. That scales nicely and we don't have to touch `cmake-tool/helpers/rootserver.cmake` for any new platforms.

axel-h (changes requested):

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,28 @@ function(DeclareRootserver rootservername)
...
-                        "\\1"
-                        host
-                        "${CROSS_COMPILER_PREFIX}"
-                )
-                get_filename_component(host ${host} NAME)
                 file(GLOB_RECURSE deps)
                 add_custom_command(
-                    OUTPUT "${CMAKE_BINARY_DIR}/bbl/bbl"
-                    COMMAND mkdir -p ${CMAKE_BINARY_DIR}/bbl
+                    OUTPUT "${OPENSBI_PATH}/build/platform/sifive/fu540/firmware/fw_payload.elf"
```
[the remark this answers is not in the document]
alistair23: That then fails as `elf_target_file` is `$<TARGET_FILE:elfloader>`

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -22,7 +22,7 @@ _start:
   addi  gp, gp, %pcrel_lo(1b)
 .option pop
 
-/* BBL starts all HARTs and sends them to this entry point. */
+/* OpenSBI starts all HARTs and sends them to this entry point. */
```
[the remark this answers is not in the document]
alistair23: It's possible that there might one day be an implementation that doesn't, so I think we should leave this as is.

alistair23:
I just pushed an update that uses the config from `<kernel>/src/plat/${you-name-it}/config.cmake`

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,33 @@ function(DeclareRootserver rootservername)
             else()
                 set(march rv64imafdc)
             endif()
-            if(UseRiscVBBL)
-                # Package up our final elf image into the Berkeley boot loader.
+            if(UseRiscVOpenSBI)
+                # Package up our final elf image into OpenSBI.
                 # The host string is extracted from the cross compiler setting
```
axel-h, line 102: I think we can drop the comment "_The host string is extracted from the cross compiler setting minus the trailing '-'_" from the code also, because this was something for Proxy-PK and all the CMake code block seem to be gone bow. There is not longer a "host" variable anywhere

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,33 @@ function(DeclareRootserver rootservername)
...
+                if("${KernelOpenSBIPlatform}" STREQUAL "")
+                    message(FATAL_ERROR "KernelOpenSBIPlatform not set.")
+                endif()
+
                 file(GLOB_RECURSE deps)
                 add_custom_command(
-                    OUTPUT "${CMAKE_BINARY_DIR}/bbl/bbl"
-                    COMMAND mkdir -p ${CMAKE_BINARY_DIR}/bbl
+                    OUTPUT
+                        "${OPENSBI_PATH}/build/platform/${KernelOpenSBIPlatform}/firmware/fw_payload.elf"
```
axel-h, line 115: I unhappy with having twice "${OPENSBI_PATH}/build/platform/${KernelOpenSBIPlatform}/firmware/fw_payload.elf" in the CMake code. This should be an internal helper variable, as this make maintenance easier.

Can we do it this way:
```
set(
    OPENSBI_FW_PAYLOAD_ELF 
    "${OPENSBI_PATH}/build/platform/${KernelOpenSBIPlatform}/firmware/fw_payload.elf"
)
add_custom_command(
   OUTPUT "${OPENSBI_FW_PAYLOAD_ELF}"
   ....
   DEPENDS "${elf_target_file}"
)
# overwrite elf_target_file, it's the the OpenSBI payload ELF file now
set(elf_target_file "${OPENSBI_FW_PAYLOAD_ELF}")
```

axel-h:
@alistair23: thanks for the changes. Can you address my last two remarks and then from my side this if good to be merged

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,31 @@ function(DeclareRootserver rootservername)
...
                 file(GLOB_RECURSE deps)
+                set(
+                    OPENSBI_FW_PAYLOAD_ELF
+                    "${OPENSBI_PATH}/build/platform/${KernelOpenSBIPlatform}/firmware/fw_payload.elf"
+                )
                 add_custom_command(
-                    OUTPUT "${CMAKE_BINARY_DIR}/bbl/bbl"
-                    COMMAND mkdir -p ${CMAKE_BINARY_DIR}/bbl
+                    OUTPUT "${OPENSBI_FW_PAYLOAD_ELF}"
+                    COMMAND mkdir -p "${CMAKE_BINARY_DIR}/opensbi"
```
axel-h, line 117: Why are we still doing `mkdir -p "${CMAKE_BINARY_DIR}/opensbi"` here actually? Seem we can also drop this, because we do a `cd "${OPENSBI_PATH}` now to run the OpenSBI build

axel-h (changes requested):

alistair23:
I have addressed all of your comments, this should be good to go.

axel-h (approved):

alistair23:
What's the plan for getting this merged? They all need to be merged at the same time

axel-h:
@alistair23: thanks, that looks perfect now. I got told that due to the holiday season, not much merging may actually happen until the first week of January.

cmake-tool/helpers/rootserver.cmake
```diff
@@ -101,37 +97,30 @@ function(DeclareRootserver rootservername)
...
-                        cd ${CMAKE_BINARY_DIR}/bbl && ${BBL_PATH}/configure
-                        --quiet
-                        --host=${host}
-                        --with-arch=${march}
-                        --with-payload=${elf_target_file}
-                            && make -s clean && make -s > /dev/null
-                    DEPENDS ${elf_target_file} elfloader ${USES_TERMINAL_DEBUG}
+                        cd "${OPENSBI_PATH}" && ${CMAKE_OBJCOPY} -O binary "${elf_target_file}"
+                        "payload" && make -s clean && CROSS_COMPILE=${CROSS_COMPILER_PREFIX} make
+                        PLATFORM="${KernelOpenSBIPlatform}" FW_PAYLOAD_PATH=payload
```
alexf91, line 120: It's not necessary to build in the source directory.
The OpenSBI Makefile defines a variable "O" to set the build directory and the payload binary can be created in the build directory.

axel-h: great, that will be very useful. Can you propose a Patch for this?

alexf91: https://github.com/seL4-drone-project/seL4_tools/commit/6f8814ee0950e2ebf989de16e2b801ce71eec4e9

axel-h: @alistair23: I'd leave it to you to have a look at the changed from @alexf91, check if the work for you also then merge them into this PR. Otherwise, we'll see that we get your PR finally merged first and then have @alexf91 make a PR on top of this.

axel-h:
@szhuang: it shows you still have change request on this PR. Can you check if they are addressed and close them? So we can get this merged.

szhuang (approved):

szhuang:
> @szhuang: it shows you still have change request on this PR. Can you check if they are addressed and close them? So we can get this merged.

All addressed, I've no objections.

cmake-tool/helpers/rootserver.cmake
alistair23: I have changed this to use the "O" variable.

This should be ready to merge now.

alexf91 (commented):
* Missing double-quote around `${OPENSBI_PATH}/build/payload"`
* Wrong FW_PAYLOAD_PATH for the final make call
* OpenSBI ist still built in the source tree

You can just merge https://github.com/seL4/seL4_tools/pull/46#discussion_r550980010, where the OpenSBI source directory isn't touched.

alistair23:
Ah, I messed up the quotes when fixing the style.

I have fixed it so it actually doesn't build in the OpenSBI directory, I also included your SoB line in the patch, let me know if that isn't ok.

alistair23:
Ping!

Jesse-Millwood:
I wonder if it would be useful to take out all bootloader specific stuff here and move it to its own repo or separate cmake file. This way more bootloaders could be supported. As it is now, the bbl support seems to be in an odd place in the build process. It might be better for a bootloader cmake function to operate on the final image rather than pull in the otherwise final image, wrap it up, and rename it to what the final image would have been.

Apave24:
I agree with @Jesse-Millwood's statement. In the interest of modularity and expandability, the rootserver.cmake would ideally contain only generic bootloader references and/or function calls and the bootloader build steps would be included based on the selected bootloader and platform. Ideally a separate repo would be used as I think bootloader integration could end up being quite extensive and a separate repo would provide more flexibility, however, separate cmake files are a good start.

axel-h:
So, practical question: Should this get merge so we have something better available and then somebody has to factor this out. Or should this PR be rejected. I'd really to avoid a situation where practical improvement get blocked due to discussion about how thing could be, but this might not happen for a long time.

Jesse-Millwood:
I'd vote merging now and moving the bootloader stuff out later maybe after more discussion.

lsf37:
> I'd vote merging now and moving the bootloader stuff out later maybe after more discussion.

I'd also prefer that.

alistair23:
Agreed. Can this be merged then?

alistair23:
Any thoughts on merging this?

oliver-wm:
@ssrg-bamboo `test`

ssrg-bamboo:
Hello, I'm a bot! I'll bring this PR into Trustworthy Systems and run some tests

ssrg-bamboo:
All the tests we ran have passed! Nice job!

PR #54, opened 2021-03-08
ElfLoader improvements, better 32/64-bit support

axel-h:
* better 32/64-bit support
* general code quality improvements
* needed for https://github.com/seL4/util_libs/pull/73
* this PR is split out from https://github.com/seL4/seL4_tools/pull/58 to allow merging some things

### DIFF
--- a/elfloader-tool/include/types.h
+++ b/elfloader-tool/include/types.h
@@ -6,28 +6,122 @@
 
 #pragma once
 
-typedef signed char int8_t;
-typedef short int16_t;
-typedef int int32_t;
-typedef long long int64_t;
+/*------------------------------------------------------------------------------
+ * helper macro that ensure the passed macro gets evaluated first before the
+ * concatenation happens. The concat macros even work when a parameter is
+ * defined to be empty,
+ */
+#define _macro_make_string_helper(x)    # x
+#define _macro_make_string(x)           _macro_make_string_helper(x)
 
-typedef unsigned char uint8_t;
-typedef unsigned short uint16_t;
-typedef unsigned int uint32_t;
-typedef unsigned long long uint64_t;
+#define _macro_concat2_helper(x,y)      x ## y
+#define _macro_concat2(x,y)             _macro_concat2_helper(x,y)
+
+#define _macro_concat3_helper(x,y,z)    x ## y ## z
+#define _macro_concat3(x,y,z)           _macro_concat3_helper(x,y,z)
+
+
+/*------------------------------------------------------------------------------
+ * define basic fixed size integer types
+ */
+typedef signed char     int8_t;
+typedef unsigned char   uint8_t;
+
+typedef signed short    int16_t;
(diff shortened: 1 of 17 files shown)


elfloader-tool/include/types.h
```diff
@@ -6,28 +6,80 @@
...
+
+
+#if defined(__KERNEL_32__)
+
+#define _int64_type         long long
+#define _int64_type_fmt     ll  // for printf() formatting and integer suffix
+
+#elif defined(__KERNEL_64__)
+
+#define _int64_type         long long
```
kent-mcleod, line 38: Can this not be a long here?

axel-h: That would have been my next question about this PR. So far I have just taken what was in the code before, which is `long long` for both 32-bit and 64-bit. Note that this is the same than what the kernels does. We can change this easily and just use `long` on 64-bit systems. **If nobody shouts, I will make this change**, as it is much more aligned with what our runtime libs do in user space. 
That leaves the question, if we should also try this change in the kernel then also. There there any reason (besides lazyness maybe) that is uses `long long` for u64 on 64-bit and not just `long`?

kent-mcleod: I'm not sure why it was long long historically, if the ABI for the compiler is LP64 and not LLP64 then I agree we should be consistent (https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models)

kent-mcleod (approved):
Nice,
I recently found that you can get quite significant startup time improvements by using an arch specific implementation of memcpy and changing `unpack_elf_to_paddr` to only memset(0) the program non-LOAD parts of the program headers. It was almost 100x faster for me on a cortex-a8.  So maybe that's what we could look to improve next

axel-h:
I have another round of improvement for RISC-V also once this is merged.

axel-h:
I've separate the RISC-V specific improvement to https://github.com/seL4/seL4_tools/pull/58 now. So hopefully I don't have to touch this PR any longer and you can put it into the merge queue.

kent-mcleod:
@axel-h: Are you happy removing the restyle commit from this PR also (context: https://github.com/seL4/util_libs/pull/73#discussion_r596510184)?  If we decide we want to change the style rules for particular libraries we can change the style tool configuration and then make the change explicitly.

kent-mcleod:
> @axel-h: Are you happy removing the restyle commit from this PR also (context: [seL4/util_libs#73 (comment)](https://github.com/seL4/util_libs/pull/73#discussion_r596510184))? If we decide we want to change the style rules for particular libraries we can change the style tool configuration and then make the change explicitly.

I see that removing this style commit would require a lot of work resolving merge conflicts in #58. So in this case I can withdraw my request for the commit to be removed as it's still correct style and all internal to the elfloader project.  But this does mean that anyone else that has changes derived from the main branch would have to resolve merge conflicts themselves as a result of the style change.

Issue #47, opened 2020-10-27
ElfLoader should use monitor.S only if CONFIG_ARM_MONITOR_HOOK is set


axel-h:
The file monitor.S should only be compiled in (or it's content take into account), when CONFIG_ARM_MONITOR_HOOK is set. Otherwise the content should be ignored, so any accidental referenced make the build fail. 

We ran into this issue when adding a new imx6 board platform, where we had to define something there to VECTOR_BASE even if this is not used anywhere because CONFIG_ARM_MONITOR_HOOK is not supported there.

axel-h:
See also https://github.com/seL4/seL4_tools/pull/53

kent-mcleod:
Resolved by https://github.com/seL4/seL4_tools/pull/56

PR #56, opened 2021-03-08
exclude monitor.S is CONFIG_ARM_MONITOR_HOOK is not active

axel-h:
address https://github.com/seL4/seL4_tools/issues/47 and exclude monitor.S is CONFIG_ARM_MONITOR_HOOK is not active


kent-mcleod (approved):

axel-h:
can we merge this?

PR #58, opened 2021-03-11
RISC-V SBI/ElfLoader improvements

axel-h:
RISC-V SBI/ElfLoader improvements.
Note that this patch builds on top of https://github.com/seL4/seL4_tools/pull/54, so it contains all commits from there.

### DIFF
--- a/elfloader-tool/include/binaries/elf/elf32.h
+++ b/elfloader-tool/include/binaries/elf/elf32.h
@@ -126,19 +126,19 @@ static inline uint32_t elf32_getSectionFlags(
 
 char const *elf32_getSectionName(
     struct Elf32_Header const *file,
-    int i);
+    unsigned int i);
 
 uint32_t elf32_getSectionSize(
     struct Elf32_Header const *file,
-    int i);
+    unsigned int i);
 
 uint32_t elf32_getSectionAddr(
     struct Elf32_Header const *elfFile,
-    int i);
+    unsigned int i);
 
 void const *elf32_getSection(
     struct Elf32_Header const *file,
-    int i);
+    unsigned int i);
 
 void const *elf32_getSectionNamed(
     struct Elf32_Header const *file,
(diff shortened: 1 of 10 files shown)


oliver-wm:
This looks fine to me. This brings up the style issues addressed [here](https://github.com/seL4/util_libs/pull/73). That needs to be changed or at least have an issue somewhere proposing why we want to adopt this and updates to the style guide accordingly. Also this is a very large PR and could be split up into more specific smaller ones to help ease the review process. Luckily most of this is just cosmetic :)

elfloader-tool/CMakeLists.txt
```diff
@@ -281,7 +281,6 @@ if(DEFINED platform_yaml)
         VERBATIM
         DEPENDS ${platform_yaml} ${PLATFORM_SIFT}
     )
-    set_property(SOURCE src/common.c PROPERTY OBJECT_DEPENDS ${PLATFORM_INFO_H})
```
kent-mcleod, line 284: I don't know how CMake can find this dependency automatically.  It seems like CMake is doing too much magic here.  But according to the docs this does seem able to be removed. 
"This property need not be used to specify the dependency of a source file on a generated header file that it includes. Although the property was originally introduced for this purpose, it is no longer necessary. If the generated header file is created by a custom command in the same target as the source file, the automatic dependency scanning process will recognize the dependency"

kent-mcleod (approved):
A lot of great improvements in here!

axel-h:
I will rework and split this up. This is basically my working branch, that why so much stuff has accumulated here.

elfloader-tool/CMakeLists.txt
axel-h: CMake can't find it automatically. When the compiler has run the first time, dep files are created.This is where ninja gets them from and know how to rebuils thing. It just gets tricky when there is no ninja rule that created them. Then CMake has to create them when the build workspace is initialized. But that create would be hard-coded then anyway in the CMake scripts. So the smarter solution in my opinion is always, have a rule that can create things on demand by ninja. Avoid leaving this in CMake, so CMake does not even have to be invoked at all for a re-build.

axel-h:
As requested, I'm splitting this up. So reducing this to firs two commits that add error codes that bubble up. I will file more PRs then when this is merged.

axel-h:
Any blocker to merge the 3 commits? For the rest I will file new PRs so reviewing/merging is easier. Intention of all this is improving the general ElfLoader code and make it easier to add more experimental code locally for experiments with early boot things.

oliver-wm (approved):

PR #59, opened 2021-03-22
[risc-V] switch to HART 1 on boot

wom-bat:
The seL4 kernel and libraries currentl assume they are running
on the first HART in the system.

OpenSBI and U-Boot pick a HART at random to start on.
This patch uses an SBI call to start HART 1 and run the elfloader on it if
at boot time a different HART has been chosen.

This is a first straw-man example of how it might be done.
Signed-off-by: Peter Chubb <peter.chubb@data61.csiro.au>

### DIFF
--- a/elfloader-tool/include/arch-riscv/sbi.h
+++ b/elfloader-tool/include/arch-riscv/sbi.h
@@ -31,6 +31,25 @@
     a0;                         \
 })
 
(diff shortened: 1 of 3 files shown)


oliver-wm:
Merged in 7fdfe5402a2ade4a11687b34d8532361d13d2352

PR #68, opened 2021-04-13
risc-v: switch to hart1

oliver-wm:
seL4 and librarys assume booting on hart1 while
Opensbi and U-boot pick a hart at random.
This patch uses an SBI call to start the elfloader on hart1
if a different hart has been choses.
Credits go to Peter Chubb for the straw-man example.

Signed-off-by: Oliver Scott <Oliver.Scott@data61.csiro.au>

### DIFF
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -6,14 +6,22 @@
 
 #include <autoconf.h>
 #include <elfloader/gen_config.h>
+
 .extern main
-.global _start
 .extern __global_pointer$
 .extern elfloader_stack_alloc
+.extern hsm_exists
 
 #define BIT(n) (1 << (n))
+#define SBI_HSM_BASE 0x48534DULL
+#define SBI_HSM_BASE_HART_START 0
+#define SBI_EXT_BASE 0x10
+#define SBI_EXT_BASE_PROBE_EXT 3
+
 .section ".text.start"
 
+/* OpenSBI starts all HARTs and sends them to this entry point. */
+.global _start
 _start:
 
 .option push
@@ -22,7 +30,48 @@ _start:
   addi  gp, gp, %pcrel_lo(1b)
 .option pop
 
-/* OpenSBI starts all HARTs and sends them to this entry point. */
+/* a0 should have hart id, store it in temp so as not to clobber from HSM calls */
+  mv t3, a0
+
+/* Check Heart State Management extension exists so we can use that to switch harts
+   if we are not hart 1, otherwise rely on the legacy extension. */
+
+/* sbi_probe_extension: Returns 0 if the given SBI extension ID (EID) is not available,
+   or an extension-specific non-zero value if it is available.*/
+
+  jal spi_probe_extension
+  mv a0, t3
+  li s1, 1
+  blt x1, s1, _start1
+
+/* Update bool to tell boot code HSM exists */
+  la t1, hsm_exists
+  li t2, 1
+  amoadd.w t1, t2, (t1)
+
+  mv a0, t3
+  li s0, CONFIG_FIRST_HART_ID
+  beq  a0, s0, _start1
+
+/* start hart 1*/
+hsm_switch_hart:
+  li a7, SBI_HSM_BASE
+  li a6, SBI_HSM_BASE_HART_START
+  li a0, 1       // hart id to start
+  la a1, _start1 // where to start new hart
+  li a2, 0       // passed in a1 when new hart starts, this is the logical hart_id
+  ecall
+
+  j secondary_harts
+
+_start1:
+
+.option push
+.option norelax
+1:auipc gp, %pcrel_hi(__global_pointer$)
+  addi  gp, gp, %pcrel_lo(1b)
+.option pop
+
   li s0, CONFIG_FIRST_HART_ID
   bne  a0, s0, secondary_harts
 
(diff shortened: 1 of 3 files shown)


oliver-wm:
revised from [59]( https://github.com/seL4/seL4_tools/pull/59)

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -24,7 +28,7 @@ _start:
 
 /* OpenSBI starts all HARTs and sends them to this entry point. */
   li s0, CONFIG_FIRST_HART_ID
-  bne  a0, s0, secondary_harts
+  bne  a0, s0, switch_heart
```
nomadeel, line 31:
```suggestion
  bne  a0, s0, switch_hart
```

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -41,6 +45,14 @@ bootstack_secondary_cores:
 
 .text
 
+switch_heart:
```
nomadeel, line 48:
```suggestion
switch_hart:
```

nomadeel (approved):
Suggestions are mostly nits, looks good to me otherwise.

elfloader-tool/src/arch-riscv/crt0.S
oliver-wm: :+1:

elfloader-tool/src/arch-riscv/crt0.S
oliver-wm: :+1:

kent-mcleod:
What is supposed to happen if the bootloader only supports the legacy SBI extension?

oliver-wm:
> What is supposed to happen if the bootloader only supports the legacy SBI extension?

Should we even support the legacy SBI extension?

kent-mcleod:
> Should we even support the legacy SBI extension?

The only SBI calls that we currently use are legacy SBI extensions: https://github.com/seL4/seL4/blob/master/include/arch/riscv/arch/sbi.h#L38

This allows the kernel to still be used with older SBI implementations such as the BBL.

> I can add a check to call the existing code based on the SBI version or something like this.

I think that is a good idea.

oliver-wm:
Sorry I should have been a bit more clear.

> This allows the kernel to still be used with older SBI implementations such as the BBL

Do we want to support BBL? I suppose it wont hurt, but I don't see why people would be using it over more mature alternatives. I don't have that much context though.

That is the only reason they exist, but are now depreciated in favor of the listed exts [here] (https://github.com/riscv/riscv-sbi-doc/blob/master/riscv-sbi.adoc#legacy-extensions-eids-0x00-0x0f). I suppose our code could more accurately reflect that each call is a separate extension, if I understand it correctly.

axel-h:
I'd propose that you can select the SBI implementation. That is how I am using this at the moment. Depending what SBI you have, this defnes how the commands are actually implemented. Then we don'*t have to worry about the details now. I can make a PR with my changes for SBI selection.

oliver-wm:
Ok @kent-mcleod I've added the check

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -41,6 +48,21 @@ bootstack_secondary_cores:
 
 .text
 
+switch_hart:
+/* Check HSM extension exists so we can use that to switch
+   otherwise rely on the legacy extension */
+  jal spi_probe_extension
+  li s0, 1
+  blt x1, s0, secondary_harts
```
yyshen, line 56: please add a bit info about what value  is returned by spi_probe_extension and why the value is compared with 1.

elfloader-tool/src/arch-riscv/console.c
```diff
@@ -5,7 +5,7 @@
  */
 
 #include <elfloader_common.h>
-#include "sbi.h"
+#include <sbi.h>
```
yyshen, line 8: This should be in another commit?

elfloader-tool/src/arch-riscv/crt0.S
yyshen: so if the extension is not supported,  the current hart is not hart1, and SMP is not enabled, the hart just spins. Is that the expected results?

oliver-wm: Yes that's right. BBL worked like this. If using OPENSBI this code path shouldn't be used.

elfloader-tool/src/arch-riscv/console.c
kent-mcleod: Or this change can be dropped as it doesn't affect the rest of the PR I believe.

oliver-wm: Yeah thought I would just fix it up. Not sure if it was intended.

kent-mcleod:
Are you able to create some riscv SMP test configurations for the CI?

DieracDelta:
Conceptually, how does this work with SMP? Suppose we're lucky and openSBI starts elfloader with hart 1/CONFIG_FIRST_HART_ID. Is elfloader supposed to start up all secondary cores and have them spin until sel4 gives a IPI? Or is that delegated to the seL4 kernel?

oliver-wm:
> Conceptually, how does this work with SMP? Suppose we're lucky and openSBI starts elfloader with hart 1/CONFIG_FIRST_HART_ID. Is elfloader supposed to start up all secondary cores and have them spin until sel4 gives a IPI? Or is that delegated to the seL4 kernel?

Yes, elfloader starts the other cores now once main has initialised from hart 1.

> A couple of details about why I'm asking this: I'm trying to get SMP running on the hifive board with openSBI. My test application is sel4test with SMP support enabled (`NUM_CORES` is 4). When I have elfloader spin up only hart 1, sel4 hangs. I'm guessing sel4 is waiting for the other harts (which I have explicitly stopped) to register an IPI and start. On the other hand, when I have elfloader spin up all harts (with only hart 1 being passed through to elf loader's `main`), `getCurrentCPUIndex` comes out to 4051. I'm guessing the per-hart kernel stack pointer isn't initialized correctly as this is how `getCurrentCPUIndex` obtains the index. Perhaps someone can shed some light on what the boot control flow is supposed to look between sel4 and elfloader when SMP support is enabled.

I've just fixed a bug with the registers, some parameters were added to the elfloader code and the kernel code was never updated. So this should fix your issue. See https://github.com/seL4/seL4/commit/baad619db316b2085944a5bddd9096d7452b9599

PR #73, opened 2021-05-13
Reset CNTVOFF on all arm generic timer platforms

xurtis:
When the kernel runs in EL1 on a platform with the arm generic timer the virtual counter is used. MCS can only correctly schedule in the first 3/8ths of representable time so must ensure that the offset is reset for the virtual timer while the bootloader is still in EL2 as EL1 cannot change the offset used to generate the virtual counter value.

### DIFF
--- a/elfloader-tool/src/plat/apq8064/platform_init.c
+++ b/elfloader-tool/src/plat/apq8064/platform_init.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <autoconf.h>
+#include <mode/arm_generic_timer.h>
+
+/* Reset the virtual offset for the platform timer to 0 */
+void platform_init(void)
+{
+    reset_cntvoff();
+}
+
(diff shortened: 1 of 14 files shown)


sylGauthier (approved):
All good

elfloader-tool/src/plat/apq8064/platform_init.c
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <autoconf.h>
+#include <mode/arm_generic_timer.h>
+
+/* Reset the virtual offset for the platform timer to 0 */
+void platform_init(void)
+{
+    reset_cntvoff();
```
kent-mcleod, line 13: If someone wanted to deduplicate this code in the future, then adding the `/timer` device to the `seL4,elfloader-devices` property in the device trees for the relevant platforms and adding a generic-timer ELFLOADER_DRIVER that resets the cntvoff would probably be the way to do it.

xurtis: It's not clear what the best way to modify this might be given that the timer needs to be initialized once per for from the given core. Maybe introducing a timer interface to reset the timer on each core?

PR #74, opened 2021-05-18
CMakeLists: Fix image start addr generation call

nomadeel:
The 'echo' calls was not generating the header files properly, adding in
whitespace between the '>' symbols allows CMake to properly parse the
arguments and invoke the calls correctly.


axel-h:
good catch, makes me wonder actually if it might be better to use the cmake file output functions here then

axel-h (approved):

oliver-wm (approved):

axel-h:
Why did this pass the tests actually, are we missing a test configuration system that uses this build path?

PR #75, opened 2021-05-20
riscv: Reposition .rodata to after .bss

nomadeel:
Previously having .rodata before .data was causing the 32-bit spike
simulation on QEMU to freeze up before entering the elfloader.
Positioning .rodata just after .bss as was the case before 6531c8e9
fixes the issue.


axel-h:
I wonder, can't we solve the root issue here instead of working on the symptom - why is RISC-V 32-bit really failing here? Having a contiguous blob with code + ro-data that can e put in memory seems a better idea to me than splitting up code and ro-data with a bss in between

nomadeel:
Well, besides from fixing the symptom, I haven't tried looking further beyond this issue. As far as I can tell right now without more investigation is that the RISC-V 32-bit is freezing up when trying to print out information, so I suspect that something's going wrong when the CPU is trying to load data. Maybe.

nomadeel:
I don't have the time to fix it fully at the moment and I'll like to get this merged so CI gets going again. I'll put up a Github issue so that we remember this.

axel-h:
Please add at least a big comment in the linker file also, so it is absolutely clear why this layout has been chosen. Can we make this conditional for 32-bit spike and leave it for all the other RISC-V platforms?

nomadeel:
Sure, will do.

oliver-wm (approved):

xurtis:
This sounds potentially as though some code is executing with the `gp` register set to an incorrect value. My suspicion would be tha there is something that is in the data section being accessed that is accessed without issue with pc-relative offset but fails for gp-relative offset when the symbol isn't within range of the pc.

xurtis (approved):

elfloader-tool/src/arch-riscv/linker.lds
```diff
@@ -4,6 +4,13 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+/*
+ * NOTE: For RISC-V 32-bit, having the .rodata section before the .data section causes the elfloader to
+ *       freeze up. Thus, for 32-bit we move the .rodata section to after the .bss section as previously
+ *       before some refactoring. The 64-bit version does not seem to be affect by this issue and uses thus
```
axel-h, line 10: typo: "affected"

elfloader-tool/src/arch-riscv/linker.lds
```diff
@@ -4,6 +4,13 @@
  * SPDX-License-Identifier: GPL-2.0-only
  */
 
+/*
+ * NOTE: For RISC-V 32-bit, having the .rodata section before the .data section causes the elfloader to
```
axel-h, line 8: for readability, please break the line after 80 chars. Especially in comments this should be followed, as centuries of typography have shown that the average human prefers even less 70 chars per line.

PR #77, opened 2021-05-25
CMake: Work around a weird CMake bug

nomadeel:
There's a weird interaction with a similar command in the capdl-loader
in the capDL repository which causes CMake to ignore the
platform_sift.py call in this repository. Merging the two
`add_custom_command` calls tells CMake to not ignore it.


nomadeel:
The issue can be reproduced with this `init-build.sh` invocation on a camkes manifest: `../init-build.sh -DCAMKES_APP=dma-example -DPLATFORM=sabre -DSIMULATION=ON -DCAmkESCapDLStaticAlloc=TRUE - DCAmkESLargeFramePromotion=ON`. The static allocation flag is the most important flag in the reproducible command.

sylGauthier (approved):
I feel like this is yet another incremental hack that makes things more and more complicated and confusing but if that works then ship it I guess...

nomadeel:
Well, I'll be happy to revisit it in the future, once things cool down in CI. I think I might even have some time to do it while waiting for CI to pass.

PR #82, opened 2021-06-03
riscv: Add clear_bss to crt0 if image is binary

nomadeel:
Binary images do not have the luxury of having a proper loader load
the elfloader into memory correctly and zero out the BSS, so we have to
do it ourselves.

### DIFF
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -67,6 +67,7 @@ hsm_switch_hart:
   li a2, 0       // passed in a1 when new hart starts, this is the logical hart_id
   ecall
 
+  mv a0, t3 // Copy the ID of the HART again, since we've clobbered a0
   j secondary_harts
 
 _start1:
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -30,8 +30,8 @@ _start:
   addi  gp, gp, %pcrel_lo(1b)
 .option pop
 
-/* a0 should have hart id, store it in temp so as not to clobber from HSM calls */
-  mv t3, a0
+/* a0 should have hart id, store it in s0 so as not to clobber from HSM calls */
+  mv s0, a0
 
 #ifdef CONFIG_IMAGE_BINARY
 /* Clear the BSS before we get to do anything more specific */
@@ -45,7 +45,7 @@ _start:
    or an extension-specific non-zero value if it is available.*/
 
   jal spi_probe_extension
-  mv a0, t3
+  mv a0, s0
   li s1, 1
   blt x1, s1, _start1
 
@@ -54,7 +54,7 @@ _start:
   li t2, 1
   amoadd.w t1, t2, (t1)
 
-  mv a0, t3
+  mv a0, s0
   li s0, CONFIG_FIRST_HART_ID
   beq  a0, s0, _start1
 
@@ -67,7 +67,7 @@ hsm_switch_hart:
   li a2, 0       // passed in a1 when new hart starts, this is the logical hart_id
   ecall
 
-  mv a0, t3 // Copy the ID of the HART again, since we've clobbered a0
+  mv a0, s0 // Copy the ID of the HART again, since we've clobbered a0
   j secondary_harts
 
 _start1:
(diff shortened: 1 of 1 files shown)


oliver-wm:
Where is clear bss?

nomadeel:
https://github.com/seL4/seL4_tools/blob/master/elfloader-tool/src/common.c#L33

nomadeel:
Added an extra commit for a fix that I found as well.

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -62,6 +67,7 @@ hsm_switch_hart:
   li a2, 0       // passed in a1 when new hart starts, this is the logical hart_id
   ecall
 
+  mv a0, t3 // Copy the ID of the HART again, since we've clobbered a0
```
kent-mcleod, line 70: How is t3 still guaranteed to have the hart id in it? Isn't t3 caller-save so can be clobbered by the ecall?

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -33,6 +33,11 @@ _start:
 /* a0 should have hart id, store it in temp so as not to clobber from HSM calls */
   mv t3, a0
 
+#ifdef CONFIG_IMAGE_BINARY
+/* Clear the BSS before we get to do anything more specific */
```
axel-h, line 37: * I consider this comment a bit too vague. TheC compiler expects bss to be filled with zeros, but for any low level hand written asm code this is not necessarily true.
* Looking at the clear_bss() implementation in C, that is just a native zero-byte-by-byte loop it might be more elegant to inline this code here and then have full control over the register usage. Then we can preserved registers easily and don't have to worry about a calling convention.
* There is a core register for the hard ID, so why to we actual care about preserving this values? We could simply read this registers when needed. Otherwise the code need some explanation why this can't be used and we need to preserve a value passed by another boot stage.

elfloader-tool/src/arch-riscv/crt0.S
nomadeel: Perhaps? I'm not well versed in RISC-V assembly to know that this is the case. From what I've tested on the Hifive, this is works fine even when the boot HART is not HART 1.

elfloader-tool/src/arch-riscv/crt0.S
nomadeel: > I consider this comment a bit too vague. TheC compiler expects bss to be filled with zeros, but for any low level hand written asm code this is not necessarily true. Furthermore, you can write (early boot) C code that does not does not use the bss, so clearing bss is not necessarily needed if the early boot code is clearly defined.

The early C boot code in SMP requires the use of a variable in BSS which is why we clear the BSS very early on. We can optimise this so that only the main HART runs it, but synchronisation with the other cores is a little difficult.

> Looking at the clear_bss() implementation in C, that is just a native zero-byte-by-byte loop it might be more elegant to inline this code here and then have full control over the register usage. Then we can preserved registers easily and don't have to worry about a calling convention.

Er, sure. I just wasn't comfortable writing RISC-V assembly and there just happened to be a `clear_bss()` shared C implementation.

> There is a core register for the hard ID, so why to we actual care about preserving this values? We could simply read this registers when needed. Otherwise the code need some explanation why this can't be used and we need to preserve a value passed by another boot stage.

I'm not sure what this is referring to. Is this directed to the other commit?

> We don't have to zero the bss on all core and should not do this. Only the primary core should do this and the asm should contain the code to sort this out.

Re above to the comments above about optimisation and assembly.

elfloader-tool/src/arch-riscv/crt0.S
axel-h: See https://github.com/riscv/riscv-elf-psabi-doc/blob/master/riscv-elf.md#register-convention, should use s0 (x8) then instead of t3 (x28)

nomadeel: Cool, I'll change that.

nomadeel:
Added a commit to use s0 instead of t3 for the HART ID.

nomadeel:
Can I review for this, I'm close to getting a release done and this is one of the blockers.

kent-mcleod (approved):

PR #83, opened 2021-06-06
EFL-Loader code cleanup and small improvements

axel-h:
Code cleanup and small improvements.

PR #86, opened 2021-06-09
elfloader: refactor code to improve maintainability

axel-h:
Thanks for merging https://github.com/seL4/seL4_tools/pull/83, and here is another round of code cleanup. It also improved the output a bit to show what the ELF loading does. This turned out to be useful for debugging to see where things end up in memory during platform porting. For day-to-day usage with a focus on running booting system images this might be a bit too verbose, but I'd leave changing this to feedback on this PR. Adding a log level concept might be a follow-up task then.

### DIFF
--- a/elfloader-tool/src/common.c
+++ b/elfloader-tool/src/common.c
@@ -29,6 +29,8 @@
 #include <platform_info.h> // this provides memory_region
 #endif
 
+#define KEEP_HEADERS_SIZE BIT(PAGE_BITS)
+
 extern char _bss[];
 extern char _bss_end[];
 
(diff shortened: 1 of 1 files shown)


axel-h:
Can we merge this?

kent-mcleod:
@axel-h, is it possible to reorganize these changes to make them easier to review? The way that the changes are currently presented in the diff viewer look as if there are > 700 lines of changes.  I'm assuming that this change isn't changing any functionality other than the added logging that you mentioned, but this is hard to review without going through every line.

axel-h:
I can try to break this down in several commit with smaller incremental changes. May take a few day. But basically, this is a big rewrite of the code.

kent-mcleod:
Making this a draft because I'm assuming @axel-h is still planning to break it up into several commits

kent-mcleod:
@axel-h, can this be closed? It can be reopened or recreated in the future, but at this stage it's unlikely to be considered as a single commit rewrite.

axel-h:
I've tried to split this up, but this takes more time in the end to make meaningful commits then. 
I'm closing this now.

PR #88, opened 2021-06-11
CMake: pass PLATFORM for OpenSBI clean step

axel-h:
Setting `PLATFORM` to a well defined value when running the clean step avoids running into conflicts where it happens to be defined as environment variable already with a value OpenSBI does not support.

See also https://github.com/riscv/opensbi/issues/217 that I have filed as potential issue there.


axel-h:
Can we merge this?

axel-h:
Can we merge this?

kent-mcleod (approved):

PR #90, opened 2021-06-19
rootserver.cmake: Set opensbi ABI and ARCH config

kent-mcleod:
The opensbi Makefile supports setting PLATFORM_RISCV_ISA and
PLATFORM_RISCV_ABI to specify the march and mabi flags that will be
used. By default the build now uses rv$(PLATFORM_RISCV_XLEN)imafdc for
the march and either ilp32d or lp64d for mabi as this is more commonly
supported by debian (and others) distribution cross compilers
(gcc-riscv64-unknown-elf)


kent-mcleod:
This makes it possible to use Debian riscv compilers to be used

lsf37 (approved):

PR #93, opened 2021-06-28
Sannum/rpi3 fixups

sannum:
A series of fixups for directly booting seL4 and seL4_test binary images on the Raspberry pi3 in both aarch32 and aarch64 modes as well as succesfully passing the dtb from the bootloader to the elfloader as this is currently not working.

The resulting sel4test-driver-image-arm-bcm2837 binary images should be directly bootable using the common bootloader.bin/start.elf and the following config.txt files:
```
enable_uart=1 
kernel=sel4test-driver-image-arm-bcm2837
kernel_address=0x[IMAGE_START_ADDR]
```
and
```
enable_uart=1 
kernel=sel4test-driver-image-arm-bcm2837
kernel_address=0x[IMAGE_START_ADDR]
arm_64bit=1
```
for aarch32 and aarch64 builds respectively.
Currently it seems the kernel_address must be manually configured to match the shoehorn generated IMAGE_START_ADDR which can be found with
```
aarch64-linux-gnu-objdump -t elfloader/elfloader | grep _text
```
as described in elfloader-tool/README.md

This also means that u-boot is not required as per https://docs.sel4.systems/Hardware/Rpi3.html.

### DIFF
--- a/elfloader-tool/src/arch-arm/sys_boot.c
+++ b/elfloader-tool/src/arch-arm/sys_boot.c
@@ -116,7 +116,7 @@ void main(UNUSED void *arg)
     print_cpuid();
     printf("  paddr=[%p..%p]\n", _text, (uintptr_t)_end - 1);
 
-#if defined(CONFIG_IMAGE_UIMAGE)
+#if defined(CONFIG_IMAGE_UIMAGE) || defined(CONFIG_IMAGE_BINARY)
 
     /* U-Boot passes a DTB. Ancient bootloaders may pass atags. When booting via
      * bootelf argc is NULL.
--- a/cmake-tool/helpers/application_settings.cmake
+++ b/cmake-tool/helpers/application_settings.cmake
@@ -12,7 +12,7 @@ include_guard(GLOBAL)
 function(ApplyData61ElfLoaderSettings kernel_platform kernel_sel4_arch)
     set(
         binary_list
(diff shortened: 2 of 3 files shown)


elfloader-tool/src/arch-arm/sys_boot.c
```diff
@@ -133,7 +133,7 @@ void main(UNUSED void *arg)
 #endif
 
     if (bootloader_dtb) {
-        printf("  dtb=%p\n", dtb);
+        printf("  dtb=%p\n", bootloader_dtb);
```
axel-h, line 136: good catch

kent-mcleod (approved):
Thanks for tracking down these fixes! I think normally the dtb is embedded into the elf binary by the build system, so this error with overwriting the dtb arg from the previous stage loader has existed for quite a while!

lsf37:
I'm not sure how, but this PR seems to have broken the elfloader boot for the IMX8MM board in 32 bit mode.

The run before in sel4test is this one: https://github.com/seL4/seL4/actions/runs/13512859723/job/37816472319 (on commit  0d12ec69 of sel4_tools)

The run after has this PR and fails on all imx8mm 32 bit sessions: https://github.com/seL4/seL4/actions/runs/13513743245/job/37760208988

kent-mcleod:
I'm taking a look

PR #95, opened 2021-07-08
Add ODroid C4 support

xurtis:
Adds the ODroid C4 as a valid device.

### DIFF
--- a/elfloader-tool/src/plat/odroidc4/platform_init.c
+++ b/elfloader-tool/src/plat/odroidc4/platform_init.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2021, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: GPL-2.0-only
+ */
+
+#include <autoconf.h>
+#include <mode/arm_generic_timer.h>
+
+/* Reset the virtual offset for the platform timer to 0 */
+void platform_init(void)
+{
+    reset_cntvoff();
+}
+
+#if CONFIG_MAX_NUM_NODES > 1
+void non_boot_init(void)
+{
+    reset_cntvoff();
+}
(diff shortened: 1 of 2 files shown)


elfloader-tool/src/plat/odroidc4/platform_init.c
```diff
@@ -0,0 +1,21 @@
...
+#include <autoconf.h>
+#include <mode/arm_generic_timer.h>
+
+/* Reset the virtual offset for the platform timer to 0 */
+void platform_init(void)
+{
+    reset_cntvoff();
+}
+
+#if CONFIG_MAX_NUM_NODES > 1
```
axel-h, line 16: There is the alias `ENABLE_SMP_SUPPORT` that is also in use, is there a policy when to use what?

axel-h: On a 2nd look, seems this is only used in the kernel, but not in the ELF loader. So ignore that comment above.

xurtis: I think `ENABLE_SMP_SUPPORT` is only defined inside the kernel.

kent-mcleod: It used to be, but then the kernel's private config file was moved into libsel4: https://github.com/seL4/seL4/blob/master/libsel4/include/sel4/config.h#L96

axel-h: Could be a thing for later then: https://github.com/seL4/seL4_tools/issues/96

nomadeel (approved):

nomadeel:
Looks good to me, the `CONFIG_MAX_NUM_NODES` could be resolved in a different PR instead.

PR #97, opened 2021-07-22
elfloader: Use file(GENERATE) for static headers

kent-mcleod:
When generating header files, the "\\n" sequence doesn't generate a
newline character unless -e is given to echo.

### DIFF
--- a/elfloader-tool/CMakeLists.txt
+++ b/elfloader-tool/CMakeLists.txt
@@ -269,15 +269,16 @@ set(PLATFORM_INFO_H "${PLATFORM_HEADER_DIR}/platform_info.h")
 set(IMAGE_START_ADDR_H "${PLATFORM_HEADER_DIR}/image_start_addr.h")
 
 if(NOT "${IMAGE_START_ADDR}" STREQUAL "")
-    add_custom_command(
-        OUTPUT "${PLATFORM_INFO_H}" "${IMAGE_START_ADDR_H}"
-        COMMAND
-            echo "#pragma once\\n\/* no platform YAML file available */" > "${PLATFORM_INFO_H}"
-        COMMAND
-            echo "#pragma once\\n#define IMAGE_START_ADDR ${IMAGE_START_ADDR}" >
-            "${IMAGE_START_ADDR_H}"
-        VERBATIM
-    )
+    # Generate static header files.  Their timestamps will change only if
+    # their contents have changed on subsequent CMake reruns.
+    file(GENERATE OUTPUT ${PLATFORM_INFO_H} CONTENT "
+#pragma once
+/* no platform YAML file available */
+")
+    file(GENERATE OUTPUT ${IMAGE_START_ADDR_H} CONTENT "
+#pragma once
+#define IMAGE_START_ADDR ${IMAGE_START_ADDR}
+")
 
 elseif(NOT DEFINED platform_yaml)
 


elfloader-tool/CMakeLists.txt
```diff
@@ -272,9 +272,9 @@ if(NOT "${IMAGE_START_ADDR}" STREQUAL "")
     add_custom_command(
         OUTPUT "${PLATFORM_INFO_H}" "${IMAGE_START_ADDR_H}"
         COMMAND
-            echo "#pragma once\\n\/* no platform YAML file available */" > "${PLATFORM_INFO_H}"
+            echo -e "#pragma once\\n\/* no platform YAML file available */" > "${PLATFORM_INFO_H}"
```
lsf37, line 275:
```suggestion
            echo -e "#pragma once\\n/* no platform YAML file available */" > "${PLATFORM_INFO_H}"
```
I think the extra backslash after the `\\n` is causing a failure now.

lsf37: Nice to see that the tests are useful :-)

axel-h:
good catch

elfloader-tool/CMakeLists.txt
kent-mcleod: Different combinations of shells and echo programs have different behaviors.

This is what I currently have:
```
diff --git a/elfloader-tool/CMakeLists.txt b/elfloader-tool/CMakeLists.txt
index 8ab64d6..56e5a90 100644
--- a/elfloader-tool/CMakeLists.txt
+++ b/elfloader-tool/CMakeLists.txt
@@ -272,10 +272,9 @@ if(NOT "${IMAGE_START_ADDR}" STREQUAL "")
     add_custom_command(
         OUTPUT "${PLATFORM_INFO_H}" "${IMAGE_START_ADDR_H}"
         COMMAND
-            echo -e "#pragma once\\n\/* no platform YAML file available */" > "${PLATFORM_INFO_H}"
+            bash -c "echo -e \"#pragma once\\n\/* no platform YAML file available */\" > ${PLATFORM_INFO_H}"
         COMMAND
-            echo -e "#pragma once\\n#define IMAGE_START_ADDR ${IMAGE_START_ADDR}" >
-            "${IMAGE_START_ADDR_H}"
+            bash -c "echo -e \"#pragma once\\n#define IMAGE_START_ADDR ${IMAGE_START_ADDR}\" > ${IMAGE_START_ADDR_H}"
         VERBATIM
     )
 
```

kent-mcleod: This part of the scripts used to use file(Write). Is anyone against changing it back? (https://github.com/seL4/seL4_tools/pull/74#issuecomment-842856262)

axel-h: The problem with with CMake's `file(WRITE ...)` was, that is cannot be used with `add_custom_command()`. So it has to be done once when CMake is initialized. The current solution can create this dynamically when rebuilding and changes are detected. I could not come up with a way to do this using `file(WRITE ...)` If echo makes problem, we could try `printf()`there. Or an inline python script ;)

lsf37: The easiest way if `echo -e` has different behaviour on different shells is probably to use two separate `echo` commands for the two lines. The backslash at `\/*` is safe to remove in any case, though. If interpreted correctly it's just `/` anyway.

axel-h: I tried to get it in one line back then to make it atomic.

kent-mcleod: > I tried to get it in one line back then to make it atomic.

It doesn't need to be atomic because the build step itself is atomic.

> The problem with with CMake's file(WRITE ...) was, that is cannot be used with add_custom_command()

Yes, but in the original code, the content being generated by file(WRITE) doesn't change dynamically. Changing the value of `IMAGE_START_ADDR` requires rerunning CMake which regenerates the file.

kent-mcleod:
I rewrote this commit to instead use `file(GENERATE)` as it avoids the complex character escaping from encoding the file contents in a shell command.

axel-h:
Seem this solves the current issue you face, so let's merge this then. When I find some more time, I will look into creating a solution that uses templates for these files and then we can generate them dynamically, but only overwrite them when there is a change. This would keep the option that `IMAGE_START_ADDR` can be changed and there is no need to re-run CMake, which was the motivation why I did this in the first place.

axel-h (approved):

kent-mcleod:
> Seem this solves the current issue you face, so let's merge this then. When I find some more time, I will look into creating a solution that uses templates for these files and then we can generate them dynamically, but only overwrite them when there is a change. This would keep the option that `IMAGE_START_ADDR` can be changed and there is no need to re-run CMake, which was the motivation why I did this in the first place.

I don't understand what situation you are referring to here. `IMAGE_START_ADDR` at this point is a cmake variable. So the only way it can be changed is by rerunning cmake. In your current implementation changing IMAGE_START_ADDR also requires rerunning cmake.

PR #103, opened 2021-08-24
Pass a1 as received from OpenSBI to main's second argument

### LINKED ISSUE #102: On a risc-v build, the second argument received from OpenSBI seems to be getting clobbered
The value of `bootloader_dtb` in `main`, arch-riscv/boot.c:290 does not correspond to the arg1 from OpenSBI.
The patch below seems to fix the problem.

NOTE: tested with `-DElfloaderIncludeDtb=OFF`

```
diff --git a/elfloader-tool/src/arch-riscv/crt0.S b/elfloader-tool/src/arch-riscv/crt0.S
index 40ab09e..28a218c 100644
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -32,6 +32,7 @@ _start:
 
 /* a0 should have hart id, store it in s0 so as not to clobber from HSM calls */
   mv s0, a0
+  mv s2, a1
 
 #ifdef CONFIG_IMAGE_BINARY
 /* Clear the BSS before we get to do anything more specific */
@@ -83,6 +84,7 @@ _start1:
 
   la sp, (elfloader_stack_alloc + BIT(12))
 
+  mv a1, s2
   la s0, main
   jr s0
 ```

### PR DESCRIPTION
**Pull request by mpolitzer**
closes #102

### DIFF
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -30,8 +30,10 @@ _start:
   addi  gp, gp, %pcrel_lo(1b)
 .option pop
 
-/* a0 should have hart id, store it in s0 so as not to clobber from HSM calls */
+/* a0 should have hart id, store it in s0 so as not to clobber from HSM calls
+   a1 should have the dtb, store it in s2 so as not to clobber it as well */
   mv s0, a0
+  mv s2, a1
 
 #ifdef CONFIG_IMAGE_BINARY
 /* Clear the BSS before we get to do anything more specific */
@@ -83,6 +85,7 @@ _start1:
 
   la sp, (elfloader_stack_alloc + BIT(12))
 
+  mv a1, s2 // restore dtb to main's second argument
   la s0, main
   jr s0
 

### CODE REVIEW

**Comment by mpolitzer** [comment]
Just rebased to master, pelase have a look.

**Comment by lsf37** [comment]
@axel-h you're the expert on OpenSBI. Is this fine to merge?

File: elfloader-tool/src/arch-riscv/crt0.S [hunk 1]
```diff
@@ -32,6 +32,7 @@ _start:
 
 /* a0 should have hart id, store it in s0 so as not to clobber from HSM calls */
   mv s0, a0
+  mv s2, a1
```
axel-h, line 35: Good catch, I have also created https://github.com/seL4/seL4_tools/issues/108 as a follow up. that we should take the passed device tree block more seriously - at least for debugging and porting that could be quite helpful. 
Could you add a comment here what this second parameters is, similar to the comment above? That improves the maintainability of the code a bit more.

mpolitzer: sure, have a look

axel-h (approved):
Thanks. It seems that neither OpenSBI v0.8 (https://github.com/riscv/opensbi/blob/v0.8/docs/firmware/fw.md) nor v0.9 (https://github.com/riscv/opensbi/blob/v0.9/docs/firmware/fw.md) have a clear specification for this interface, https://github.com/riscv/opensbi/blob/master/docs/firmware/fw.md was update just recently to clarify what is passed. The 12.1.0 release was at v0.8 and the update to v0.9 is still open. I've bee looking through the OpenSBI sources to see the DTB has always been passed there - seem so, but I'm not fully sure. However, passing a2 transparently will not really break anything, so I think we should merge this now.

Issue #43, opened 2020-09-25
print_cpuid display incorrect data for at least Cortex-A35


bennoleslie:
The function `print_cpuid` incorrectly calculates the CPU name associated with the part number.

The current code is:

```
    if ((CPUID_PART(cpuid) & 0xf00) == 0xC00) {
        printf("Cortex-A%d ", CPUID_PART(cpuid) & 0xff);
    } else if ((CPUID_PART(cpuid) & 0xf00) == 0xD00) {
        printf("Cortex-A5%d ", CPUID_PART(cpuid) & 0xff);
    } else {
        printf("Part: 0x%03x ", CPUID_PART(cpuid));
```

However, the Cortex-A35 `part` is `0xD04`. The current logic prints that as `Cortex-A54`, which is obviously wrong.

I think the assumptions made in this printing is just incorrect. A switch-based look up table is probably most appropriate. Something like:

```
    switch (CPUID_PART(cpuid)) {
       case 0xD04:
           printf("Cortex-A35 ");
            break;
        default:
            printf("Part: 0x%03x ", CPUID_PART(cpuid));
     }
```

But, obviously, with additional switch statements for other, supported, CPUs.

axel-h:
Sounds reasonable to me. Can you make a pull request with a patch?

bennoleslie:
OK. I finally made a pull request: https://github.com/seL4/seL4_tools/pull/107

PR #107, opened 2021-08-30
Fix ARM CPUID printing

bennoleslie:
The existing print_cpuid function displays the incorrect value
for Cortex-A35. (See issue #43).

This change makes CPUID determination an explicit switch statement
rather than attempting to automatically determine it through
masking certain bits.

Note: At some level this will be a regression as it now only
prints the textual id for Cortex-A35 and no other CPU.

Improvements to list other CPUID mappings is welcome, unfortunately
the only platform I have access to for testing is Cortex-A35.

### DIFF
--- a/elfloader-tool/src/arch-arm/cpuid.c
+++ b/elfloader-tool/src/arch-arm/cpuid.c
@@ -92,11 +92,105 @@ void print_cpuid(void)
     if (CPUID_ARCH(cpuid) != CPUID_ARCH_CPUID) {
         printf("%s ", cpuid_get_arch_str(cpuid));
     }
-    if ((CPUID_PART(cpuid) & 0xf00) == 0xC00) {
-        printf("Cortex-A%d ", CPUID_PART(cpuid) & 0xff);
-    } else if ((CPUID_PART(cpuid) & 0xf00) == 0xD00) {
-        printf("Cortex-A5%d ", CPUID_PART(cpuid) & 0xff);
-    } else {
+    /* References: https://en.wikipedia.org/wiki/Comparison_of_ARMv8-A_cores */
+    switch (CPUID_PART(cpuid)) {
+    case 0xC05:
+        printf("Cortex-A5 ");
+        break;
+    case 0xC07:
+        printf("Cortex-A7 ");
+        break;
+    case 0xC08:
+        printf("Cortex-A8 ");
+        break;
+    case 0xC09:
+        printf("Cortex-A9 ");
+        break;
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

axel-h:
This patch fixes A35, but breaks a lot of the rest. There is a list in last column of the table at http://en.wikipedia.org/wiki/Comparison_of_ARMv8-A_cores that needs to be worked in here.

bennoleslie:
> This patch fixes A35, but breaks a lot of the rest. There is a list in last column of the table at http://en.wikipedia.org/wiki/Comparison_of_ARMv8-A_cores that needs to be worked in here.

Added, but obviously untested. I think they are reasonably clear so can be merged without testing all cases.

axel-h:
Thanks, to merge there are two things that should be addressed:
- All these ID are only valid if `CPUID_IMPL(cpuid) == CPUID_IMPL_ARM`. Otherwise we should just print the raw value. 
- This still lacks all the ARMv7 cores
  - 0xc05: "Cortex-A5"
  - 0xc07: "Cortex-A7"
  - 0xc08: "Cortex-A8"
  - 0xc09: "Cortex-A9"
  - 0xc0d: "Cortex-A12"
  - 0xc0f: "Cortex-A15"
  - 0xc0e: "Cortex-A17"

These things are nice to have:
- There are also a few ARMv8 cores out there that we may want to add:
  - 0xd42: "Cortex-A78AE"
  - 0xd4b: "Cortex-A78C"
- This seems to be the list for the brand new ones:
  - 0xd0c, "Neoverse N1"
  - 0xd40: "Neoverse V1"
  - 0xd46: "Cortex-A510"
  - 0xd47: "Cortex-A710"
  - 0xd48: "Cortex-X2"
  - 0xd49: "Neoverse N2"
  - 0xd4a: "Neoverse E1"

bennoleslie:
> Thanks, to merge there are two things that should be addressed:
> 
> * All these ID are only valid if `CPUID_IMPL(cpuid) == CPUID_IMPL_ARM`. Otherwise we should just print the raw value.

I don't think this needs to be addressed to merge. This is a bug (that was preexisting). I think we should create a bug in the issue tracker and address this separately.

Doing that properly would be a refactor of the part lookup into a `cpuid_get_arm_part_str(...)` function. Which is a good idea. I'd just like to keep the PRs separate.

axel-h (approved):
Look good please add the missing DCO and consider squashing the commits.

PR #117, opened 2021-09-05
elfloader/risc-v: improve OpenSBI support

axel-h:
- improve comments and variable names
- allow platforms to customize OpenSBI parameters
  - `OPENSBI_PLAT_XLEN`
  - `OPENSBI_PLAT_ISA`
  - `OPENSBI_PLAT_ABI`

### DIFF
--- a/cmake-tool/helpers/rootserver.cmake
+++ b/cmake-tool/helpers/rootserver.cmake
@@ -93,7 +93,10 @@ function(DeclareRootserver rootservername)
         set(elf_target_file $<TARGET_FILE:elfloader>)
         if(KernelArchRiscV)
             if(UseRiscVOpenSBI)
-                # Package up our final elf image into OpenSBI.
+                # When using OpenSBI, the whole system image (usually consisting
+                # of the ELF-Loader, a device tree, the kernel, and a userland)
+                # is packaged as an OpenSBI payload.
+
                 if("${CROSS_COMPILER_PREFIX}" STREQUAL "")
                     message(FATAL_ERROR "CROSS_COMPILER_PREFIX not set.")
                 endif()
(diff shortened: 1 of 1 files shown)


cmake-tool/helpers/rootserver.cmake
```diff
@@ -93,7 +93,11 @@ function(DeclareRootserver rootservername)
         set(elf_target_file $<TARGET_FILE:elfloader>)
         if(KernelArchRiscV)
             if(UseRiscVOpenSBI)
-                # Package up our final elf image into OpenSBI.
+                # When using OpenSBI, the whole system image (usually consisting
+                # of the ELF-Loader, a device tree, the kernel, and a userland)
+                # is packaged as an OpenSBI payload.
+                file(GLOB_RECURSE deps)
```
kent-mcleod, line 99: What's happening here? deps looks to be unused, and this cmake command isn't given any paths to glob and so deps should end up empty too.

axel-h: I've moved the statement back where it was. If this does nothing, we can also remove it completely then.

kent-mcleod: Thanks.It looks like it's missing the other parts of it's implementation, actually specifying the path to the opensbi sources and then adding the sources to the depends line of add_custom_command. If you wanted to add a commit to this PR adding `${BBL_PATH}` to the end of that statement, and also adding the deps arg to the ${CMAKE_BINARY_DIR}/bbl/bbl custom command then that'd still come under the PR's goal of "improve OpenSBI support". Alternatively you could ignore it, or remove it if you think it's better to assume that the opensbi implementation isn't going to be edited by anyone between builds.

axel-h: Can we do this in a separate PR? I will open a dedicated issue for this.

kent-mcleod (approved):

PR #124, opened 2021-09-28
elfloader/risc-v: align register usage

axel-h:
- It's more intuitive to use `s1` to store `a1` than using `s2` here.
- use temporary register `t0` instead of `s1` for temporary values


axel-h:
@kent-mcleod This is hand-written assembly code, so the register usage should follow a pattern. I agree that if somebody else has patches, this may make is harder to merge them - but give that I still have problem understanding what this code does (and why it does this) and my other pull requests, I'd prefer to sanitize the code step by step.

axel-h:
dropping this PR, doesn't make much sense any longer to keep this factored out from a bigger change set.

PR #125, opened 2021-10-06
rootserver.cmake: remove use of config_option

kent-mcleod:
config_option is primarily intended for generating a C header file with
config options. As no C header file is used in rootserver.cmake, it's
better to use a builtin CMake function that creates the CMake option as
expected.


axel-h:
Good point. I thing it's bet to just merge this then.

PR #130, opened 2021-11-11
elfloader: pass KernelSel4Arch to python scripts

axel-h:
The parameter is optional and defaults to 'arm', which is a semantic error when building RISC-V targets. This does not cause any issues so far, because the value is mostly unused.

This change requires https://github.com/seL4/seL4/pull/642

Test with seL4/seL4#642

### DIFF
--- a/elfloader-tool/CMakeLists.txt
+++ b/elfloader-tool/CMakeLists.txt
@@ -337,6 +337,7 @@ if(DEFINED KernelDTBPath)
             --hardware-config "${config_file}"
             --hardware-schema "${schema_file}"
             --dtb "${KernelDTBPath}"
+            --sel4arch "${KernelSel4Arch}"
         VERBATIM
         DEPENDS ${KernelDTBPath} ${config_file} ${schema_file}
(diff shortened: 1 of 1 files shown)


kent-mcleod (approved):

lsf37:
Something doesn't seem to be quite working there yet. Both arm-virt qemu simulation runs crashed (gcc and clang) at boot time: https://github.com/seL4/seL4_tools/runs/4347457932?check_suite_focus=true#step:3:444

(This is tested together with the seL4 PR)

axel-h:
I will look into this.

axel-h:
There was still something in the template that expected the `arch` property. I'm keeping the `arch` property now and add a `sel4arch` property that has the deeper architecture details. Instead of the formerly optional "--arch ..." parameter "--sel4arch ..." is a required parameter now. That should help everybody to do the transition more smoothly because a  error will show up about a missing parameter. It also keeps the naming aligned with everything else.

lsf37 (approved):

PR #132, opened 2021-11-13
improve RISC-V multi core boot

axel-h:
- add much more comments
- fix SBI HSM register naming quirk
- Ensure DTB is always passed to primary core boot
- check SBI HSM error when bringing up secondary harts
- Stop wrong boot hart via HSM if possible
- unify stack definitions for all cores
- add multicore helper functions
- Drop variable hsm_exists, pass information as parameter
- Print more log messages


yyshen:
@axel-h thanks for the changes. I just want to double check if there are any bugs fixed in the PR.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -176,17 +176,52 @@ int secondary_go = 0;
...
+        /* busy waiting loop */
+    }
+}
+
+static void release_multicore_lock(void)
+{
+    __atomic_store_n(&mutex, 0, __ATOMIC_RELEASE);
+}
+
+static void set_secondary_cores_go(void)
```
yyshen, line 192: unleash_secondary_cores?

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -107,28 +107,28 @@ hsm_switch_hart:
   mv a0, s0 /* restore a0 to hold hart ID passed by OpenSBI */
   j secondary_harts
 
-
-_start1: /* a0 must hold current hard ID passed by bootloader */
-         /* a1 must hold dtb address passed by bootloader */
+/*----------------------------------------------------------------------------*/
```
yyshen, line 110: why is this needed?

elfloader-tool/src/arch-riscv/boot.c
yyshen, line 1: so this basically has no code logic changes? In this case, maybe we should state it in the commit message clearly.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -220,6 +220,7 @@ static int run_elfloader(UNUSED word_t hart_id, void *bootloader_dtb)
     }
 
 #if CONFIG_MAX_NUM_NODES > 1
+
```
yyshen: why are the additional empty lines?

Issue #84, opened 2021-06-07
asm code calls clear_bss without a valid stack pointer


axel-h:
In https://github.com/seL4/seL4_tools/pull/82 we call `clear_bss` from asm code now when `CONFIG_IMAGE_BINARY` is enabled - but there is no stack set up. This is a bit dangerous, since C code always assumes there is stack. It may work if the function is simple enough so C does not create a stack frame.

axel-h:
This will be fixed when https://github.com/seL4/seL4_tools/pull/135 is merged, as a proper stack pointer is set up then.

kent-mcleod:
Fixed by https://github.com/seL4/seL4_tools/pull/136

Issue #76, opened 2021-05-24
RISC-V linker issues for 32-bit simulation


nomadeel:
As a result from #75, we discovered that having a contiguous blob of code + ro-data presents some problems for 32-bit RISC-V simulation on QEMU. The simulation in question manages to start up the elfloader only for it to freeze up near the beginning when it's printing information about the HART that it is running on.

kent-mcleod:
It's highly likely that this issue is caused by the bss section not being initialized to 0 causing some issue with setup.
- The spike platforms aren't configured to set CONFIG_IMAGE_BINARY so clear_bss wouldn't be called.
- The opensbi compilation converts the elfloader elf into a binary image which would remove the .bss section from the end of the file as it's allocated but has default contents
- When the image is initialized by the loader, the memory isn't zeroed and may have different values.
- Rearranging the linker script to put the .rodata sections after the bss cause the bss sections to become contents within the binary file and would then be initialized by the loading.

PR #135, opened 2022-02-01
Allow elfloader to put kernel images directly after firmware on RISCV

kent-mcleod:
opensbi puts the elfloader in a fixed location right after it in memory which is where we normally wish to place the kernel with the elfloader higher in memory.  This allows the kernel to immediately reclaim the elfloader memory and slightly reduces memory fragmentation by allowing the kernel reserved memory to be located next to the opensbi reserved memory and all memory after free for application use. 

Test with: https://github.com/seL4/seL4/pull/759


lsf37:
I'm not deep enough in the RISCV details to review this one, but I am in favour of the general idea.

elfloader-tool/src/common.c
```diff
@@ -618,3 +623,41 @@ WEAK void platform_init(void)
...
+ */
+char * fixup_image_base(void) {
+    char *load_base = _start;
+    char *target_base = (char *) IMAGE_START_ADDR;
+    if (load_base == target_base) {
+        /* already in the right place, just keep booting */
+        return (char *)0;
+    }
+
+    /* Check that the current image location doesn't overlap with the
```
yyshen, line 646: we could copy backwards if they overlap, just not sure if that is necessary.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -89,11 +91,6 @@ static int map_kernel_window(struct image_info *kernel_info)
 
     /* Map the elfloader into the new address space */
 
-    if (!IS_ALIGNED((uintptr_t)_text, PT_LEVEL_2_BITS)) {
```
yyshen, line 92: could you explain why the check existed previously and why it can be removed a bit?

elfloader-tool/src/common.c
kent-mcleod: Actually the memmove implementation in the elfloader can handle this. I originally thought that it was just the same implementation of memcpy.  I can fix this up to handle overlap apart from the code segment apart so that it's like the original assembly (which I originally thought was not properly handling overlap).

elfloader-tool/src/arch-riscv/boot.c
kent-mcleod: Because the kernel also could only handle setting up the kernel mappings if the image was on a page boundary and because the elfloader was always placed on a page boundary by the bbl/opensbi implementation. The rest of map_kernel_window was specialized to these assumptions and so had the checks to catch when there was divergence. Now that divergence can be handled by the code, the checks should be able to be removed.

yyshen (approved):

cmake-tool/helpers/application_settings.cmake
```diff
@@ -49,9 +49,6 @@ function(ApplyData61ElfLoaderSettings kernel_platform kernel_sel4_arch)
     if(KernelPlatformZynqmp AND KernelSel4ArchAarch32)
         set(IMAGE_START_ADDR 0x8000000 CACHE INTERNAL "" FORCE)
     endif()
-    if(KernelPlatformSpike AND KernelSel4ArchRiscV32)
-        set(IMAGE_START_ADDR 0x80400000 CACHE INTERNAL "" FORCE)
```
yyshen, line 53: this can be removed because of the DTS overlay change in another PR?

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -54,10 +54,25 @@ _start:
   mv s0, a0 /* preserve a0 (hart id) in s0 */
   mv s2, a1 /* preserve a1 (dtb) in s2 */
 
-#ifdef CONFIG_IMAGE_BINARY
+  /* Set a stack so that we can call into C functions */
+  la sp, (elfloader_stack_alloc + BIT(12))
```
axel-h, line 58: seem this als addresses https://github.com/seL4/seL4_tools/issues/84 then.

elfloader-tool/src/common.c
axel-h: Isn't the issue when there is overlapping, there is chance that the code that is currently executing is not overwritten and thus the result might be undefined? So this must be detected and avoided. Same for overwriting the stack/global in bss

elfloader-tool/src/arch-arm/64/crt0.S
```diff
@@ -41,70 +41,6 @@ BEGIN_FUNC(_start)
     b       main
 END_FUNC(_start)
 
-#ifdef CONFIG_IMAGE_BINARY
-BEGIN_FUNC(fixup_image_base)
-    stp     x29, x30, [sp, #-16]!
-    mov     x29, sp
-    ldr     x0, =IMAGE_START_ADDR
```
axel-h, line 48: Seem there is no need to have `#include <image_start_addr.h>` in this file any longer then when this usage here is gone.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -19,7 +19,11 @@
 #define PT_LEVEL_2 2
 
 #define PT_LEVEL_1_BITS 30
+#if __riscv_xlen == 32
+#define PT_LEVEL_2_BITS 22
```
axel-h, line 23: Did this work in the past for RV32 just because we have been lucky the addresses in use were "choosen well"?

kent-mcleod: Yes.  And because the elfloader and kernel elf fit inside a single page mapping. The subsequent PPN in the mapping code would have been wrong.

elfloader-tool/src/arch-riscv/crt0.S
kent-mcleod: I made https://github.com/seL4/seL4_tools/pull/136 to explicitly fix the early boot issues.

elfloader-tool/src/common.c
kent-mcleod: I've fixed this up to only check for collision up until _archive_start.

kent-mcleod:
These changes also allow the issue underlying https://github.com/seL4/seL4_tools/issues/76 to be resolved.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -59,7 +59,7 @@ unsigned long l2pt[PTES_PER_PT] __attribute__((aligned(4096)));
 unsigned long l2pt_elf[PTES_PER_PT] __attribute__((aligned(4096)));
 #endif
 
-char elfloader_stack_alloc[BIT(CONFIG_KERNEL_STACK_BITS)];
+char elfloader_stack_alloc[BIT(CONFIG_KERNEL_STACK_BITS)] __attribute__((section(".data"))) ;
```
axel-h, line 62: Can you add a comment here that explains why this is in the data section? Could be the same as in the commit comment.

kent-mcleod:
This relocation needs to consider the DTB passed through from opensbi when doing relocations otherwise it could get overwritten.  This is an outstanding issue on Arm (https://sel4.atlassian.net/browse/SELFOUR-2138)

cmake-tool/helpers/application_settings.cmake
kent-mcleod: Yea, it was being ignored by opensbi before and wasn't having any effect.

kent-mcleod:
> This relocation needs to consider the DTB passed through from opensbi when doing relocations otherwise it could get overwritten. This is an outstanding issue on Arm (https://sel4.atlassian.net/browse/SELFOUR-2138)

This is now handled by the most recent commit.

elfloader-tool/CMakeLists.txt
```diff
@@ -344,6 +344,8 @@ if(DEFINED KernelDTBPath)
     set_property(SOURCE src/drivers/driver.c PROPERTY OBJECT_DEPENDS ${DEVICES_GEN_H})
 endif()
 
+set_property(SOURCE src/common.c PROPERTY OBJECT_DEPENDS ${IMAGE_START_ADDR_H})
```
axel-h, line 347: Is this really necessary? With this PR `#include <image_start_addr.h>` is added to `common.c`, so it should be part of the normal dependency generation.

elfloader-tool/src/common.c
```diff
@@ -618,3 +623,67 @@ WEAK void platform_init(void)
...
+        }
+        if ((fdt >= target_base && fdt < target_end) ||
+            (fdt_end >= target_base && fdt_end < target_end)) {
+            abort();
+        }
+    }
+
+
+    /* Perform the move and clean/invalidate caches if necessary */
+    char *ret = memmove(target_base, load_base, image_size);
```
axel-h, line 682: Drop `ret` and use `target_base` instead, so it's easier to understand what this does. We could assert ro abort if `memmove()` does not return `target_base`, because that indicated something is terribly broken.

elfloader-tool/src/arch-arm/64/crt0.S
```diff
@@ -26,12 +23,13 @@ BEGIN_FUNC(_start)
      * Binary images may not be loaded in the correct location.
      * Try and move ourselves so we're in the right place.
      */
+    mov     x0, #0 /* Arg 0 of fixup_image_base is NULL */
```
axel-h, line 26: Isn't there a possibility on ARM also that we get passed a DTB here, and thus we have to check that it's not overwritten?

elfloader-tool/src/common.c
```diff
@@ -618,3 +623,67 @@ WEAK void platform_init(void)
...
+/*
+ * Moves the elfloader to its expected start address in memory.
+ * This is for when the elfloader needs to be in a specific location
+ * to ensure that it doesn't occupy memory that it needs to load the
+ * kernel and user applications into.
+ */
+char *fixup_image_base(char const *fdt)
+{
+    char *load_base = _start;
+    char *target_base = (char *) IMAGE_START_ADDR;
```
axel-h, line 641:
```suggestion
    char *target_base = (char *)IMAGE_START_ADDR;
```

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -57,7 +59,10 @@ unsigned long l2pt[PTES_PER_PT] __attribute__((aligned(4096)));
 unsigned long l2pt_elf[PTES_PER_PT] __attribute__((aligned(4096)));
 #endif
 
-char elfloader_stack_alloc[BIT(CONFIG_KERNEL_STACK_BITS)];
+/* This stack cannot go in the .bss section because it's already in use when the
+ * .bss section is zeroed.
```
axel-h, line 63: I wonder, don't we have the same problem on ARM also then? Seem implementing `clear_bss` in assembly, so it does not use a stack is the cleanest way to avoid running into issues here.

elfloader-tool/CMakeLists.txt
kent-mcleod: I'll take it out and test if it picks it up automatically

kent-mcleod: CMake does detect that the custom command that generates this header file is a dependency of the c file. 
I checked that it works both ways:
- It won't build common.c for the first time until after the header has been generated,
- if the header is updated it will recompile the comon.c file. 
So it is adding it explicitly to the dependency graph as a dependency of common.c automatically.

The part I never understand is how it knows to wait to build common.c the first time until it's generated the header file. Because it's only by building common.c that it gets the make depends output from the compiler listing the header file as a dependency. It must just always build all custom commands within the same target before the target or something

elfloader-tool/src/arch-riscv/boot.c
kent-mcleod: Arm already places the stack outside of the bss section. In any case, I've changed this commit to use the same approach as arm. This way the stack is still not allocated into the elf file, and it's not zeroed when the rest of the bss is zeroed.

elfloader-tool/src/common.c
```diff
@@ -618,3 +623,48 @@ WEAK void platform_init(void)
...
+#endif
+
+/*
+ * Moves the elfloader to its expected start address in memory.
+ * This is for when the elfloader needs to be in a specific location
+ * to ensure that it doesn't occupy memory that it needs to load the
+ * kernel and user applications into.
+ */
+char * fixup_image_base(void) {
+    char *load_base = _start;
```
kent-mcleod, line 640: This will only work if this assignment uses a pc relative address. This code isn't loaded compiled with any position independent flags. So this might be a bit too dodgy.  Maybe it should take the real base address as an argument.
```
    80a1c752:   ffffe797                auipc   a5,0xffffe
    80a1c756:   8ae78793                add     a5,a5,-1874 # 80a1a000 <_start>
```

elfloader-tool/src/arch-arm/64/crt0.S
kent-mcleod: CONFIG_IMAGE_BINARY doesn't support passing a DTB at this stage.

PR #136, opened 2022-02-04
elfloader,RISC-V: Fix early boot issues

kent-mcleod:
- Set a valid stack before calling any C function so that memory and
  registers don't get corrupted.
- Keep track of the DTB address so that it's still in a1 when main is
  called even if we needed to switch to a different HART.
- If the HSM extention isn't present, then try and call main on the
  current HART.
- Suspend HARTs that exceed CONFIG_MAX_NUM_NODES.

### DIFF
--- a/elfloader-tool/src/arch-riscv/crt0.S
+++ b/elfloader-tool/src/arch-riscv/crt0.S
@@ -17,6 +17,7 @@
 /* SBI commands */
 #define SBI_HSM_BASE 0x48534DULL
 #define SBI_HSM_BASE_HART_START 0
+#define SBI_HSM_BASE_HART_STOP 1
 #define SBI_EXT_BASE 0x10
 #define SBI_EXT_BASE_PROBE_EXT 3
 
@@ -54,6 +55,9 @@ _start:
   mv s0, a0 /* preserve a0 (hart id) in s0 */
   mv s2, a1 /* preserve a1 (dtb) in s2 */
 
+  /* Attach the stack to sp before calling any C functions */
+  la sp, (elfloader_stack_alloc + BIT(12))
+
 #ifdef CONFIG_IMAGE_BINARY
 /* Clear the BSS before we get to do anything more specific */
   jal clear_bss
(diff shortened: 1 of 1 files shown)


kent-mcleod:
Fixes https://github.com/seL4/seL4_tools/issues/84

lsf37:
Thanks for separating that out, much easier to review.

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -100,16 +106,17 @@ hsm_switch_hart:
 
 
 _start1: /* a0 must hold current hard ID passed by bootloader */
-
+         /* a1 must hold dtb address passed by bootloader */
```
axel-h, line 109: `a1` is overwritten with the value from `s2` line 125, this must be removed.

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -153,6 +160,13 @@ secondary_harts:
   la s0, secondary_entry
   jr s0
 #endif
+/* If we get here then the HSM extension exists and the current
```
axel-h, line 163: could put `hsm_suspend_hart` into the "#if CONFIG_MAX_NUM_NODES > 1" block, becuase it will only be called on SMP

elfloader-tool/src/arch-riscv/crt0.S
kent-mcleod: Removed

elfloader-tool/src/arch-riscv/crt0.S
kent-mcleod: It should still be needed to fall through when `CONFIG_MAX_NUM_NODES == 1` and the boot core is not `CONFIG_FIRST_HART_ID`

axel-h (approved):
Look good to me to be merged

elfloader-tool/src/arch-riscv/crt0.S
```diff
@@ -88,8 +94,8 @@ hsm_switch_hart:
   li a7, SBI_HSM_BASE
   li a6, SBI_HSM_BASE_HART_START
   li a0, CONFIG_FIRST_HART_ID /* hart id to start */
+  mv a2, s2 /* dtb address to be passed in a1 when new hart starts is 3rd parameter */
```
lsf37, line 97: to be passed in `a1` (as the comment says) or in `a2` as the code does?

lsf37: From what I can tell, the comment is probably the one that is wrong. We seem to be calling with parameters `a0 = CONFIG_FIRST_HART_ID, a1= _start1, a2=dtb`, which would make sense.

lsf37 (approved):

elfloader-tool/src/arch-riscv/crt0.S
kent-mcleod: The third parameter of the sbi call specifies the value of a1 when the new Hart starts. So a2 holds the a1 value

lsf37: Ok, so after the call `a1` will be `dtb`. Happy with that.

lsf37:
Good to merge?

PR #142, opened 2022-03-20
Add --dry-run to simulate.py

locriacyber:


cmake-tool/simulate_scripts/simulate.py
```diff
@@ -47,9 +49,7 @@ def parse_args():
 def notice(message):
     # Don't call this without initialising `progname`.
     assert(progname)
-    sys.stderr.write("{}: {}".format(progname, message))
-    sys.stderr.flush()
-
+    print("{}: {}".format(progname, message), file=sys.stderr)
```
axel-h, line 52: Please make this change a separate commit.

cmake-tool/simulate_scripts/simulate.py
```diff
@@ -80,23 +80,25 @@ def notice(message):
...
-        msg = "QEMU failed; resetting terminal in {d} seconds".format(d=delay) \
-            + "--interrupt to abort\n"
-        notice(msg)
-    else:
-        delay = 2  # in seconds
-
-    time.sleep(delay)
-
-    subprocess.call("tput reset", shell=True)
+        print(f'GDB command: {qemu_gdbserver_command}', file=sys.stderr)
```
axel-h, line 84: Shouldn't this be printed only if `args.dry_run` is set?

cmake-tool/simulate_scripts/simulate.py
lsf37: And please don't remove the flush, the output is not necessarily on a terminal and if a script is calling this somewhere you want the error message in the logs a close to the cause as possible.

axel-h: using `printf(...., flush=true)`is also an option.

lsf37:
Hi @locriacyber, thank for the contribution. Can you please add a description to the PR and commit what you are trying to achieve and why?

Please also see <https://docs.sel4.systems/processes/git-conventions.html> for more info on the git conventions used in these repositories and <https://docs.sel4.systems/processes/contributing.html> for how to add a sign-off line to commits.

cmake-tool/helpers/check_arch_compiler.cmake
```diff
@@ -101,7 +96,7 @@ function(check_arch_gcc)
 endfunction()
 
 function(check_arch_compiler)
-    if(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
```
axel-h, line 99: Please consider making this a separate commit with a comment that explains the issue that is fixed here.

locriacyber:
I'll make another branch

cmake-tool/simulate_scripts/simulate.py
locriacyber: Nothing else is writing to stderr?

axel-h: > In anything else is writing to stderr?

Not sure I get the context of this question. This script could run in a bigger context and thus there might be more that is written to `stderr`. Maybe even within the same python interpreter session. But from my experience with various CI system (and their Web-GUIs) , it is usually a good idea to force flushes to see then output in the "right" place.

cmake-tool/simulate_scripts/simulate.py
axel-h: Yes, this is a bit unfortunate sometimes, but it survives in captured logs at least.
But I think this line is meant to be printed to give the user a hint what to do next, so it should be printed always. Thus Ignore my remark.

cmake-tool/simulate_scripts/simulate.py
lsf37: Yes, CI is exactly where log reordering happens all the time without flushing.

PR #143, opened 2022-03-21
Add --dry-run to simulate.py

locriacyber:
Previously, it's impossible to read the QEMU command because QEMU switch to another screen when run.

This option let me read the command to run QEMU without running it.

### DIFF
--- a/cmake-tool/simulate_scripts/simulate.py
+++ b/cmake-tool/simulate_scripts/simulate.py
@@ -35,6 +35,8 @@ def parse_args():
                         help="Kernel file to pass onto QEMU", default="@QEMU_SIM_KERNEL_FILE@")
     parser.add_argument('-i', '--initrd', dest='qemu_sim_initrd_file', type=str,
                         help="Initrd file to pass onto QEMU", default="@QEMU_SIM_INITRD_FILE@")
+    parser.add_argument('-n', '--dry-run', dest='dry_run', action='store_true',
+                        help="Output qemu command, but do not run it")
     parser.add_argument('--extra-qemu-args', dest='qemu_sim_extra_args', type=str,
                         help="Additional arguments to pass onto QEMU", default="@QEMU_SIM_EXTRA_ARGS@")
     parser.add_argument('--extra-cpu-opts', dest='qemu_sim_extra_cpu_opts', type=str,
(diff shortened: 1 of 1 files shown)


cmake-tool/simulate_scripts/simulate.py
```diff
@@ -47,9 +49,7 @@ def parse_args():
 def notice(message):
     # Don't call this without initialising `progname`.
     assert(progname)
-    sys.stderr.write("{}: {}".format(progname, message))
-    sys.stderr.flush()
-
+    print("{}: {}".format(progname, message), file=sys.stderr)
```
axel-h: this change should to into the other commit, it's not related to the dry-run feature

axel-h: Maybe it's easier if you swap the commits, first fix the printing and then add the dry-run

axel-h:
I like the idea of of the dry-run parameter. Improves user-friendliness and allows getting the command line without hacking into the scripts. Using `printf()` instead or `sys.stderr` also seems a bit nicer, so there is no need to use the `sys.stderr` directly an longer.

cmake-tool/simulate_scripts/simulate.py
locriacyber: I don't know how to seperate the commits...

locriacyber:
printf in python?

cmake-tool/simulate_scripts/simulate.py
axel-h: I meant separate the change set and make two commits from this instead of one.

axel-h:
> printf in python?

got me, it's `print()` there actually.

locriacyber:
Is this merge-able now?

axel-h:
The code change looks good to me.

kent-mcleod:
> Is this merge-able now?

Content-wise yes, however the DCO requirement will still need to be met.  I guess we wait for a resolution in https://github.com/seL4/sel4-tutorials/pull/89.

lsf37:
Closing this for inactivity because the DCO requirements are not met and it does not look like loriacyber is planning to change that. The DCO is not a requirement we can compromise on unless we're talking about really trivial changes.

I don't see anything wrong with the content and would be happy to take this contribution, so if anything there changes please do feel free to re-open.

Issue #145, opened 2022-04-20
elfloader-tool: MAIR attributes for aarch64 (question)


malus-brandywine:
Does anyone remember why aarch64 code doesn't define memory region "type": "Normal memory, cacheable, write-through" in MAIR_EL1 ?

</br>

Memory refresher - this is what we have now:
```
/*
 * Memory types are defined in Memory Attribute Indirection Register.
 *  - nGnRnE Device non-Gathering, non-Reordering, No Early write acknowledgement
 *  - nGnRE Unused Device non-Gathering, non-Reordering, Early write acknowledgement
 *  - GRE Unused Device Gathering, Reordering, Early write acknowledgement
 *  - NORMAL_NC Normal Memory, Inner/Outer non-cacheable
 *  - NORMAL Normal Memory, Inner/Outer Write-back non-transient, Write-allocate, Read-allocate
 * Note: These should match with contents of MAIR_EL1 register!
 */
enum mair_types {
    DEVICE_nGnRnE = 0,
    DEVICE_nGnRE = 1,
    DEVICE_GRE = 2,
    NORMAL_NC = 3,
    NORMAL = 4
};

```

kent-mcleod:
The aarch64 user API only currently exposes 2 different memory attribute options based on the `seL4_ARM_PageCacheable` option of `seL4_ARM_VMAttributes`:
- DEVICE_nGnRnE (`0b00000000`)
- NORMAL (`0b11111111`

Because of this, unless the kernel needed to create it's own mappings with different attributes, other attribute encodings weren't added to the `mair_el1` registers.

PR #146, opened 2022-04-27
elfloader/arm64: add "write-through" support

malus-brandywine:
Memory region type: "Normal memory, write-through" was added.

Signed-off-by: Nataliya Korovkina <malus.brandywine@gmail.com>

Test with: seL4/seL4#834


malus-brandywine:
Addresses issue #145

kent-mcleod (approved):

PR #149, opened 2022-06-30
CMake: remove unused variable

axel-h:
I can't see where `append` is used and the git history also does not reveal it.


kent-mcleod:
> I can't see where `append` is used and the git history also does not reveal it.

I recall it used to be used to modify the cpio command to change from create to append after the first time through the `foreach` loop.

kent-mcleod (approved):

PR #151, opened 2022-07-07
improve CPIO creation

axel-h:
- simplify check for unused parameters]
- reduce redundancy by using some helper variables
- use verbose cpio flags
- Improve maintainability for CPIO creation
- allow passing a directory also in the CPIO archive name, so root binary folder pollution can be avoided


cmake-tool/helpers/cpio.cmake
```diff
@@ -36,11 +36,11 @@ endfunction()
 # into another target
 function(MakeCPIO output_name input_files)
     cmake_parse_arguments(PARSE_ARGV 2 MAKE_CPIO "" "CPIO_SYMBOL" "DEPENDS")
-    if(NOT "${MAKE_CPIO_UNPARSED_ARGUMENTS}" STREQUAL "")
```
kent-mcleod, line 39: This pattern is repeated a lot throughout the CMake files.  I recall it's partially because `if (${MAKE_CPIO_UNPARSED_ARGUMENTS})` unexpectedly evaluates to false for almost all string values as  CMake's described behavior (https://cmake.org/cmake/help/latest/command/if.html).

Using `if(MAKE_CPIO_UNPARSED_ARGUMENTS)` is shorter and you know that this tests if the variable is defined and defined to non-false values. But it would be easy to misread something like `if(${MAKE_CPIO_UNPARSED_ARGUMENTS}` or `if("${MAKE_CPIO_UNPARSED_ARGUMENTS}"`) as testing for truth  if you aren't more familiar with the if() gotchas . In these cases, if the variable contains a string it will always evaluate to false unless the contents match the truthy string values (eg ON, TRUE, etc). So while `"${MAKE_CPIO_UNPARSED_ARGUMENTS}" STREQUAL "")` is more verbose, it's probably a good pattern to leave around for explicitly testing for a non-empty string as it doesn't rely on understanding any if() testing gotchas.

cmake-tool/helpers/cpio.cmake
```diff
@@ -44,9 +44,15 @@ function(MakeCPIO output_name input_files)
...
+    get_filename_component(cpio_lib_dir "${output_name}" DIRECTORY)
+    if(NOT cpio_lib_dir)
+        set(cpio_lib_dir "${CMAKE_CURRENT_BINARY_DIR}")
+    endif()
+
     # intermediate files
-    set(cpio_archive "archive.${output_name}.cpio")
-    set(cpio_archive_s "${output_name}.S")
+    set(cpio_archive "${cpio_lib_dir}/${cpio_lib_name}.cpio")
+    set(cpio_archive_s "${cpio_lib_dir}/${cpio_lib_name}.S")
```
kent-mcleod, line 55: What happens to existing calls to MakeCPIO that don't pass in any DIRECTORY path component?
Eg in the elfloader: `MakeCPIO(archive.o "${cpio_files}" CPIO_SYMBOL _archive_start)`, `cpio_lib_dir` would be set to "" and then the cpio_archive would be created at `/${cpio_lib_name}.cpio` I think?

cmake-tool/helpers/cpio.cmake
axel-h: I agree with all points. This is the outcame from digging into CMake usage and a long term plan to use the language as it seems intended nowadays. I can move this to a separate PR to make the intention more clear. There are more places where this could be changed to clarify what checks are non-trivial.

cmake-tool/helpers/cpio.cmake
axel-h: I will check this case again. I thought I caught all corner cases from testing the practical usage, as our internal CI runs passed.

axel-h: That case is handeled, there is a check in line 49 that handles this:
```
    if(NOT cpio_lib_dir)
        set(cpio_lib_dir "${CMAKE_CURRENT_BINARY_DIR}")
    endif()
```

kent-mcleod: Ok thanks

kent-mcleod (dismissed):

axel-h:
Rebased. Can we merge this now, do the other PRs that use this get unblocked?

cmake-tool/helpers/cpio.cmake
axel-h: I have remove the commit

axel-h:
rebased and improved the absolute/relative path handling a bit more.

kent-mcleod:
When I originally approved this, it was a much smaller change: https://github.com/seL4/seL4_tools/commit/e5981ae372a350993178b39d64f72e179ed082ea, so my approval doesn't apply to the current version of the PR.

cmake-tool/helpers/cpio.cmake
```diff
@@ -39,48 +39,100 @@ function(MakeCPIO output_name input_files)
...
-    set(
-        commands
-        "bash;-c;cpio ${cpio_reproducible_flag} --quiet --create -H newc --file=${CMAKE_CURRENT_BINARY_DIR}/archive.${output_name}.cpio;&&"
+    set(cpio_archive_creator "${output_name_abs}.sh")
+    file(
+        WRITE
+            "${cpio_archive_creator}"
+            # content
+            "#!/bin/sh\n"
+            "# auto-generated file from MakeCPIO(), do not edit\n"
```
kent-mcleod, line 77: Something like `set -e` is needed here or when the script is invoked to handle when any of the inner commands return an error.

cmake-tool/helpers/cpio.cmake
```diff
@@ -39,48 +39,100 @@ function(MakeCPIO output_name input_files)
...
+            "${archive_symbol}_end:\n"
+    )
+
+    # Re-run CMake configuration in case 'cpio_archive_creator' or
+    # 'cpio_archive_s' is deleted.
+    set_property(
+        DIRECTORY
+        APPEND
+        PROPERTY CMAKE_CONFIGURE_DEPENDS "${cpio_archive_creator}" "${cpio_archive_s}"
+    )
```
kent-mcleod, line 111: I don't consider that this property is needed unless CMake implicitly adds it for all files it generates in the build directory already.  Trying to detect and recover the build directory becoming corrupted isn't the responsibility of the build scripts.

cmake-tool/helpers/cpio.cmake
axel-h: added, thanks.

cmake-tool/helpers/cpio.cmake
axel-h: It's a convenience feature I found quite useful when experimenting with the archive creation. In general, when debugging CMake things, I often corrupt the build folder or delete files on purpose to force partial regeneration. And my expectation from proper dependency management is, that everything in the build folder should be re-generated automatically.

axel-h:
> When I originally approved this, it was a much smaller change: [e5981ae](https://github.com/seL4/seL4_tools/commit/e5981ae372a350993178b39d64f72e179ed082ea), so my approval doesn't apply to the current version of the PR.

The older approach did not work in some corner cases.

PR #152, opened 2022-10-19
cmake_tool: add zynqmp to platform_strings

chrisguikema:
Without adding the zynqmp subplatforms (zcu102, ultra96, ultra96v2), sel4test can only be compiled for the generic zynqmp platform. While this is compatible with the ZCU102, the Ultra96 platforms use a different serial port, so the sel4test logs would be unreadable.

This commit allows sel4test to be compiled with -DPLATFORM=ultra96 and -DPLATFORM=ultra96v2.

Signed-off-by: Chris Guikema <chrisguikema@dornerworks.com>

### DIFF
--- a/cmake-tool/helpers/application_settings.cmake
+++ b/cmake-tool/helpers/application_settings.cmake
@@ -162,6 +162,7 @@ function(correct_platform_strings)
         "bcm2711:rpi4"
         "exynos5:exynos5250,exynos5410,exynos5422"
         "am335x:am335x-boneblack,am335x-boneblue,am335x-bone"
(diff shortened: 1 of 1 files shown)


axel-h (approved):
Thanks, looks good to me. Turns I've been using almost the same patch locally for some time...

lsf37 (approved):

axel-h:
@chrisguikema : Now that this canse is in, I wonder if https://docs.sel4.systems/Hardware/ZCU102.html should also get an update now and say `../init-build.sh -DPLATFORM=zcu102 -DAARCH64=1` instead of `... -DPLATFORM=zynqmp`

PR #153, opened 2022-10-27
elfloader/cmake-tool support for RPi4

JoonasOnatsu:
This PR adds SMP support for RPi4 to ```elfloader```.

### DIFF
--- a/elfloader-tool/src/plat/bcm2711/smp.c
+++ b/elfloader-tool/src/plat/bcm2711/smp.c
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019, ARM Ltd
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ *
+ */
+
+#include <autoconf.h>
+#include <elfloader/gen_config.h>
+#include <types.h>
+
+#if CONFIG_MAX_NUM_NODES > 1
+#include <types.h>
+#include <elfloader.h>
+#include <armv/machine.h>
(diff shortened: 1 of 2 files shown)


Ivan-Velickovic:
I am able to run seL4test without any changes to the elfloader using the latest U-Boot, maybe older U-Boot versions did not support EFI images?

JoonasOnatsu:
> I am able to run seL4test on a RPi4 without any changes to the elfloader using the latest U-Boot, maybe older U-Boot versions did not support EFI images?

Hi Ivan, thanks for your comment. I also tested with the latest U-Boot, and it seems that you are correct and the image can be booted with 'bootefi'. I think I had something misconfigured before, and I couldn't boot the image.

I removed the commit for changing the image type.

elfloader-tool/src/plat/bcm2711/smp.c
```diff
@@ -0,0 +1,50 @@
...
+#include <armv/machine.h>
+#include <armv/smp.h>
+#include <printf.h>
+#include <abort.h>
+
+#define MAX_CORES 4
+
+extern void core_entry_head(unsigned long stack);
+
+unsigned long core_stack;
```
axel-h, line 24: If this holds an address, then we should simply say so.
```suggestion
void *core_stack;
```

kent-mcleod:
Superseded by https://github.com/seL4/seL4_tools/pull/222/

PR #154, opened 2022-11-09
cmake: support qemu-riscv-virt platform

axel-h:
Test with: seL4/seL4#818, seL4/util_libs#143, seL4/sel4test#82


cmake-tool/helpers/simulation.cmake
```diff
@@ -152,6 +152,16 @@ function(GenerateSimulateScript)
...
         SetDefaultMemSize("${QEMU_MEMORY}")
+    elseif(KernelPlatformQEMURiscVVirt)
+        set(QemuBinaryMachine "qemu-system-${KernelSel4Arch}")
+        set(sim_machine "virt")
+        # defaults for the virt machine:
+        #     aclint=off (on/off)
+        #     aia=none (none/aplic/aplic-imsic)
+        #     aia-guests=0 (VS-level AIA IMSIC pages per hart)
+        set(sim_graphic_opt "-nographic")
+        set(sim_cpu "rv${KernelWordSize}")
```
lsf37, line 163:
```suggestion
        set(sim_cpu "rv${KernelWordSize}")
        set(qemu_sim_extra_args "-bios none")
```

lsf37: (see Ivan's comment in https://github.com/seL4/seL4/pull/818)

axel-h: I think the better solution would be that we make the script support `-bios <image>` instead of `-kernel <image>` on RISC-V.

axel-h: Since the `-bios none` is used for the other targets also, I think it's easier to use this for now and merge things to have qemu-riscv-virt working. I will make another PR for `-bios <image>` then.

lsf37: :+1:

lsf37 (approved):

PR #155, opened 2022-11-26
gitlint: enable regex-style-search explicitly

axel-h:
Setting `regex-style-search=true` avoids a warning. 
```
WARNING: I1 - ignore-by-title: gitlint will be switching from using Python regex 'match' (match beginning) to
'search' (match anywhere) semantics. Please review your ignore-by-title.regex option accordingly.
To remove this warning, set general.regex-style-search=True. 
More details: https://jorisroovers.github.io/gitlint/configuration/#regex-style-search
```
So far, gitlint uses Python's `re.match()` (match from start) by default, but it will change to use the more generic `re.search()` (match anywhere) in the future. Since we prefer `re.search()`, the new behavior is enabled  explicitly.  For more details see also
https://jorisroovers.com/gitlint/configuration/#regex-style-search.

### DIFF
--- a/misc/.gitlint
+++ b/misc/.gitlint
@@ -14,6 +14,14 @@ ignore-fixup-commits=false
 # By default gitlint will ignore squash commits. Set to 'false' to disable.
 ignore-squash-commits=false
 
+# So far, gitlint uses Python's `re.match()` (match from start) by default, but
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Yes, probably time to switch this one.

PR #158, opened 2023-03-13
Reduce fudge factor from shoehorn and strip elf files before loading

kent-mcleod:
Correct the calculated memory needed to load an elf image. Using only
the p_memsz value ignores gaps between segments; instead calculate the
first+last addresses using p_vaddr & p_memsz.

As get_memory_usage in elf_sift.py now can calculate correct memory
usage from a list of ELF segments, a more accurate upper bound on
memory usage for the loaded apps is possible and the old fudge
factor is no longer needed.

Thanks to @sleffler for the patch.

### DIFF
--- a/cmake-tool/helpers/application_settings.cmake
+++ b/cmake-tool/helpers/application_settings.cmake
@@ -57,6 +57,13 @@ function(ApplyData61ElfLoaderSettings kernel_platform kernel_sel4_arch)
         endif()
         set(IMAGE_START_ADDR 0x10000000 CACHE INTERNAL "" FORCE)
     endif()
+    if(KernelPlatformSpike OR KernelPlatformQEMURiscVVirt)
+        # Spike/qemu loads elfloader to either 0x80200000 or 0x80400000
+        # But the RISC-V kernel is currently loaded to a different
+        # address than what's modelled by the shoehorn.py tool.
+        # So force the start address to one that works.
+        set(IMAGE_START_ADDR 0x81000000 CACHE INTERNAL "" FORCE)
+    endif()
     if(KernelPlatformStar64)
         set(IMAGE_START_ADDR 0x60000000 CACHE INTERNAL "" FORCE)
     endif()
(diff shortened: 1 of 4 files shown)


kent-mcleod:
There are some errors that I'll try and track down.  The Elfloader is reporting that the loaded regions are all exclusive, however there appears to be consistent hanging before executing the kernel in some cases.

```
  ELF-loader started on CPU: ARM Ltd. Cortex-A53 r0p4
    paddr=[64b000..b5d117]
  No DTB passed in from boot loader.
  Looking for DTB in CPIO archive...found at 7c9528.
  Loaded DTB from 7c9528.
     paddr=[250000..254fff]
  ELF-loading image 'kernel' to 0
    paddr=[0..24ffff]
    vaddr=[ffffff8000000000..ffffff800024ffff]
    virt_entry=ffffff8000000000
  ELF-loading image 'sel4test-driver' to 255000
    paddr=[255000..64afff]
    vaddr=[400000..7f5fff]
    virt_entry=40e548
  Boot cpu id = 0x0, index=0
  
  [[Timeout]]
  None
```

cmake-tool/helpers/application_settings.cmake
```diff
@@ -57,6 +57,13 @@ function(ApplyData61ElfLoaderSettings kernel_platform kernel_sel4_arch)
         endif()
         set(IMAGE_START_ADDR 0x10000000 CACHE INTERNAL "" FORCE)
     endif()
+    if(KernelPlatformSpike OR KernelPlatformQEMURiscVVirt)
+        # Spike loads elfloader to either 0x80200000 or 0x80400000
+        # But the RISC-V kernel is currently loaded to a differen't
```
lsf37, line 62:
```suggestion
        # But the RISC-V kernel is currently loaded to a different
```

lsf37 (approved):
Looks good. Should we close JIRA [SELFOUR-2335](https://sel4.atlassian.net/browse/SELFOUR-2335)?

PR #160, opened 2023-03-22
elfloader: factor out check_hash()

axel-h:
Cleaning up the code, so it becomes a bit more modular. Moving this out comes handy for further refactoring.
This also allows running the hash check on the image before accessing any other data from it.

The feature is enabled when compiling with `-DElfloaderHashInstructions="hash_sha"`. Is anybody using this actually?


axel-h:
Any chance to get this merged?

PR #163, opened 2023-04-30
Add dry run option to simulate script

Ivan-Velickovic:
Pretty much the change from https://github.com/seL4/seL4_tools/pull/143, I find it quite useful from time to time so it would be nice for it to be upstreamed.

### DIFF
--- a/cmake-tool/simulate_scripts/simulate.py
+++ b/cmake-tool/simulate_scripts/simulate.py
@@ -35,6 +35,8 @@ def parse_args():
                         help="Kernel file to pass onto QEMU", default="@QEMU_SIM_KERNEL_FILE@")
     parser.add_argument('-i', '--initrd', dest='qemu_sim_initrd_file', type=str,
                         help="Initrd file to pass onto QEMU", default="@QEMU_SIM_INITRD_FILE@")
+    parser.add_argument("-n", '--dry-run', dest='dry_run', action='store_true',
+                        help="Output command for QEMU (and GDB), but do not execute it")
     parser.add_argument('--extra-qemu-args', dest='qemu_sim_extra_args', type=str,
                         help="Additional arguments to pass onto QEMU", default="@QEMU_SIM_EXTRA_ARGS@")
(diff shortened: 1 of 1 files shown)


kent-mcleod:
Sure, I'd also find this useful.  Currently you can achieve this with `-b echo` which sets the binary to echo instead of qemu, but this is too hard to remember...

cmake-tool/simulate_scripts/simulate.py
```diff
@@ -80,23 +82,26 @@ def notice(message):
...
-        msg = "QEMU failed; resetting terminal in {d} seconds".format(d=delay) \
-            + "--interrupt to abort\n"
-        notice(msg)
-    else:
-        delay = 2  # in seconds
-
-    time.sleep(delay)
-
-    subprocess.call("tput reset", shell=True)
+        notice('GDB command: ' + qemu_gdbserver_command)
```
kent-mcleod, line 88: Why are these prints changed too? `qemu_gdbserver_command` is already part of qemu command I think?

Ivan-Velickovic: Yes you're right, my bad.

Ivan-Velickovic: (I've discarded that change)

kent-mcleod (approved):

cmake-tool/simulate_scripts/simulate.py
```diff
@@ -80,23 +82,25 @@ def notice(message):
...
-    notice(qemu_simulate_command)
+    notice('QEMU command: ' + qemu_simulate_command)
 
     if qemu_gdbserver_command != "":
-        notice('waiting for GDB on port 1234...')
+        if not args.dry_run:
+            notice('waiting for GDB on port 1234...')
 
-    qemu_status = subprocess.call(qemu_simulate_command, shell=True)
+    if not args.dry_run:
```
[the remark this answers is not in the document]
Ivan-Velickovic: Done, I've done `exit` since `return` is not possible.

lsf37 (approved):
Nice, will merge when the tests have passed.

Issue #164, opened 2023-05-01
Setting `UseRiscVOpenSBI` has no affect


Ivan-Velickovic:
For example, running the following with seL4test:

`../init-build.sh -DPLATFORM=hifive -DUseRiscVOpenSBI=OFF` still produces an image with OpenSBI. The option is defined in `rootserver.cmake` with `cmake_dependent_option`. The CMake documentation (https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html) says that the option should be available to the user, although I don't know what their definition of user is.

axel-h:
Readinf thge docs, I wonder if the line
```
cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON DEPENDS "KernelArchRiscV" OFF)
```
should actually be
```
cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON "KernelArchRiscV" OFF)
```

Ivan-Velickovic:
Ah I missed that, that worked. Thanks! I will make a PR.

PR #165, opened 2023-05-01
Fix `UseRiscVOpenSBI` CMake option

Ivan-Velickovic:
Closes https://github.com/seL4/seL4_tools/issues/164. Perhaps we should use `config_option` to be consistent with other CMake options? Happy to change the patch to use that instead, I don't know enough about CMake to know which is better.

### DIFF
--- a/cmake-tool/helpers/rootserver.cmake
+++ b/cmake-tool/helpers/rootserver.cmake
@@ -30,7 +30,7 @@ mark_as_advanced(TLS_ROOTSERVER)
 find_file(UIMAGE_TOOL make-uimage PATHS "${CMAKE_CURRENT_LIST_DIR}" CMAKE_FIND_ROOT_PATH_BOTH)
 mark_as_advanced(UIMAGE_TOOL)
 include(CMakeDependentOption)
-cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON DEPENDS "KernelArchRiscV" OFF)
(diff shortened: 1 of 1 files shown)


cmake-tool/helpers/rootserver.cmake
```diff
@@ -30,7 +30,7 @@ mark_as_advanced(TLS_ROOTSERVER)
 find_file(UIMAGE_TOOL make-uimage PATHS "${CMAKE_CURRENT_LIST_DIR}" CMAKE_FIND_ROOT_PATH_BOTH)
 mark_as_advanced(UIMAGE_TOOL)
 include(CMakeDependentOption)
-cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON DEPENDS "KernelArchRiscV" OFF)
+cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON "KernelArchRiscV" OFF)
```
axel-h, line 33: This is the minimal fix. 
On the bigger picture, this is the only place where we use `cmake_dependent_option()`. Everywhere else we use our extended CMake function `config_option(). So maybe we should do it here also for consistency. Das might also explain where how wrong `DEPENDS` came use.

```
config_option(
    UseRiscVOpenSBI USE_RISCV_OPENSBI
    "Use OpenSBI."
    DEFAULT ON
    DEPENDS "KernelArchRiscV"
    DEFAULT_DISABLE OFF
)
```

Also, the better option name would be something like "PackageWithOpenSBI" actually.

axel-h (approved):

kent-mcleod (approved):

PR #166, opened 2023-05-01
Rename/rework `UseRiscVOpenSBI` option

Ivan-Velickovic:
Alternative to https://github.com/seL4/seL4_tools/pull/165 that fixes https://github.com/seL4/seL4_tools/issues/164 and also makes the option more consistent with other seL4 CMake options.


axel-h:
Why did you close this? I think this change makes sense in the ling run.

axel-h:
I'm reopening this, because I think changing this improves the semantics, this is really about packaging the system with OpenSBI. 
Using OpenSBI in terms of using the SBI API is a slightly different thing, and to me separating these config options makes sense. There is not much public RISC-V hardware/simulation that does not use the standard SBI API, but there are different version. For our Chip, we have a stripped down version SBI. So, having a config option just dedicated to the SBI API makes sense.

axel-h:
@Ivan Could you rebase this

cmake-tool/helpers/rootserver.cmake
```diff
@@ -29,10 +29,16 @@ mark_as_advanced(TLS_ROOTSERVER)
 
 find_file(UIMAGE_TOOL make-uimage PATHS "${CMAKE_CURRENT_LIST_DIR}" CMAKE_FIND_ROOT_PATH_BOTH)
 mark_as_advanced(UIMAGE_TOOL)
-include(CMakeDependentOption)
-cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON DEPENDS "KernelArchRiscV" OFF)
 
-if(UseRiscVOpenSBI)
+config_option(
+    RiscVPackageOpenSBI RISCV_PACKAGE_OPENSBI
+    "Compile OpenSBI with the seL4test image as the payload"
```
axel-h, line 35:
```suggestion
    "Compile OpenSBI with the seL4 system image as the payload"
```

cmake-tool/helpers/rootserver.cmake
```diff
@@ -29,10 +29,16 @@ mark_as_advanced(TLS_ROOTSERVER)
 
 find_file(UIMAGE_TOOL make-uimage PATHS "${CMAKE_CURRENT_LIST_DIR}" CMAKE_FIND_ROOT_PATH_BOTH)
 mark_as_advanced(UIMAGE_TOOL)
-include(CMakeDependentOption)
-cmake_dependent_option(UseRiscVOpenSBI "Use OpenSBI." ON DEPENDS "KernelArchRiscV" OFF)
 
-if(UseRiscVOpenSBI)
+config_option(
+    RiscVPackageOpenSBI RISCV_PACKAGE_OPENSBI
```
axel-h, line 34: Maybe call this `RiscVPackageWithOpenSBI

kent-mcleod:
> Why did you close this? I think this change makes sense in the long run.

I closed this because it was only a proposed alternative to #165 which was already merged.

The use of `config_option` isn't correct in this context because there's no use of add_config_library() declaring a config library for this module.  `cmake_dependent_option` only declares a CMake option and doesn't require the add_config_library() infrastructure to achieve its goal.

kent-mcleod:
>  Using OpenSBI in terms of using the SBI API is a slightly different thing, and to me separating these config options makes sense. There is not much public RISC-V hardware/simulation that does not use the standard SBI API, but there are different version. For our Chip, we have a stripped down version SBI. So, having a config option just dedicated to the SBI API makes sense.

If you're saying that you want a C option for whether the kernel is using the SBI ABI then this should be declared in the kernel repo I'd propose?

axel-h:
> If you're saying that you want a C option for whether the kernel is using the SBI ABI then this should be declared in the kernel repo I'd propose?

True. It just leave name change to `RiscVPackageWithOpenSBI` to happen here.

Ivan-Velickovic:
I don't think this change is worth it given people are already using the option, someone else can feel free to take the patch and make their own PR.

PR #168, opened 2023-05-30
elfloader: Add generic-timer driver

kent-mcleod:
This driver just sets CNTVOFF_El2 to 0 if possible.
This can only be done if the elfloader is in hyp-mode, otherwise nothing
happens.

This allows us to replace most of the platform-specific code in the elfloader.

This originally came up with https://github.com/seL4/seL4_tools/pull/73#discussion_r632270450


axel-h:
I wonder, do we want keep the empty `platform_init.c` files, as they may get content in the future again? I remember a discussion somewhere else, where this was the recommended.

elfloader-tool/src/plat/fvp/platform_init.c
axel-h, line 1: Remove `#include <mode/arm_generic_timer.h>` also now, as this is not longer needed?

elfloader-tool/src/plat/exynos5/platform_init.c
axel-h, line 1: Remove `#include <mode/arm_generic_timer.h>` also now, as this is not longer needed?

elfloader-tool/src/plat/tk1/platform_init.c
axel-h, line 1: Remove `#include <mode/arm_generic_timer.h>` also now, as this is not longer needed?

kent-mcleod:
> I wonder, do we want keep the empty `platform_init.c` files, as they may get content in the future again? I remember a discussion somewhere else, where this was the recommended.

If possible we want to support platforms without having to add empty files for each one

elfloader-tool/src/plat/tx2/platform_init.c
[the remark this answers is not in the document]
kent-mcleod: fixed

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -43,6 +44,9 @@ void non_boot_main(void)
     /* Initialise any platform-specific per-core state */
     non_boot_init();
```
Indanz, line 45: Are all functions replaced with the driver `init_non_boot`? If so, `non_boot_init()` can be deleted.

elfloader-tool/src/drivers/driver.c
```diff
@@ -53,3 +53,35 @@ int initialise_devices(void)
...
+            drv->init_non_boot(dev, drv->match_table[ret].match_data);
+        }
+        drvp++;
+    }
+
+    return 0;
+
+}
+
+
```
Indanz, line 75: Convention seems one empty line between functions.

elfloader-tool/src/drivers/driver.c
```diff
@@ -53,3 +53,35 @@ int initialise_devices(void)
...
+{
+    for (unsigned int i = 0; i < ARRAY_SIZE(elfloader_devices); i++) {
+        int ret = init_device_non_boot(&elfloader_devices[i]);
+        if (ret) {
+            return ret;
+        }
+    }
+
+    return 0;
+
```
Indanz, line 86: I'd remove the empty lines around `return`, it just looks weird.

But again, if no one is checking the return value, just make this function void and keep going.

Indanz (approved):
Overall a good improvement.

I think merging this without also merging seL4/seL4#1047 will break things, so both need to be used together.

elfloader-tool/src/drivers/driver.c
```diff
@@ -53,3 +53,35 @@ int initialise_devices(void)
...
+        struct elfloader_driver *drv = *drvp;
+        int ret = table_has_match(dev->compat, drv->match_table);
+        if (ret >= 0 && drv->init_non_boot) {
+            dev->drv = drv;
+            drv->init_non_boot(dev, drv->match_table[ret].match_data);
+        }
+        drvp++;
+    }
+
+    return 0;
```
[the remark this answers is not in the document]
kent-mcleod: IF any of the setup functions return non-zero, then the loop is exited early which means that the caller needs to distinguish that this has happened. I added checks at all the callsites to handle the error cases.

elfloader-tool/src/drivers/timer/generic_timer.c
axel-h, line 1: better call this `arm_generic_timer.c`, to make it a bit more specific.

elfloader-tool/src/drivers/driver.c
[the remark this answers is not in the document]
Indanz: But there's nothing we can do, all we can hope is that the UART is setup first and print some debug message out. So can as well keep going instead of breaking the loop.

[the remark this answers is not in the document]
kent-mcleod: There's abort()

lsf37 (approved):
Nice to see so much removed code..

elfloader-tool/src/drivers/driver.c
```diff
@@ -53,3 +53,32 @@ int initialise_devices(void)
...
+{
+    struct elfloader_driver **drvp = __start__driver_list;
+
+    while (drvp < __stop__driver_list) {
+        struct elfloader_driver *drv = *drvp;
+        if (drv->init_non_boot) {
+            int ret = table_has_match(dev->compat, drv->match_table);
+            if (ret >= 0) {
+                dev->drv = drv;
+                drv->init_non_boot(dev, drv->match_table[ret].match_data);
```
Indanz, line 67: The return value of `drv->init_non_boot()` should be propagated, otherwise my old comment stands.

Indanz (approved):

PR #170, opened 2023-06-27
fix: update machine type of Raspberry Pi3b for simulation in QEMU

lopzek:
The `-machine` option should be `raspi3b`.


cmake-tool/helpers/simulation.cmake
```diff
@@ -118,7 +118,7 @@ function(GenerateSimulateScript)
     elseif(KernelPlatformRpi3 AND KernelSel4ArchAarch64)
         set(QemuBinaryMachine "qemu-system-aarch64")
         set(sim_serial_opt "-serial null -serial mon:stdio")
-        set(sim_machine "raspi3")
```
kent-mcleod, line 121: Is there a version of qemu where `raspi3` isn't supported? On my version (5.2.0) it is supported:
```
$ qemu-system-aarch64 -machine help
Supported machines are:
...
raspi3               Raspberry Pi 3B (revision 1.2) (alias of raspi3b)
raspi3b              Raspberry Pi 3B (revision 1.2)
...
```

Ivan-Velickovic: Later versions do not support `raspi3`.

Ivan-Velickovic: See https://wiki.qemu.org/ChangeLog/6.2.

lopzek:
qemu version 6.2.0 shows only support `raspi3ap` for Raspberry Pi 3A+ (revision 1.0) and
`raspi3b` for Raspberry Pi 3B (revision 1.2)

axel-h:
Looks like changing this make sense. Could you add a DCO?

lopzek:
> Looks like changing this make sense. Could you add a DCO?

DCO added

Ivan-Velickovic (approved):
The QEMU version in the supported seL4 Docker container is v5.2.0, which has `raspi3` as an alias of `raspi3b`, so this patch should have no effect there.

PR #171, opened 2023-07-13
Fix RISC-V compilation with recent GCC

wom-bat:
Default args for gcc have changed; we need to fix the args given to the compiler.


axel-h:
Seems this is https://github.com/seL4/seL4/issues/749 again.

wom-bat:
> Seems this is [seL4/seL4#749](https://github.com/seL4/seL4/issues/749) again.

Yes, but the fix there didn't fix all the  flags.

axel-h:
> Yes, but the fix there didn't fix all the flags.

You are referring to the CMake flags that are specific to OpenSBI, not some missing gcc flags, right? Because the change seem to be an exact copy of the flags.

axel-h (approved):
Looks good to me

wom-bat:
Yep.

PR #173, opened 2023-07-26
Changes to enable linking with LLVM/lld

heshamelmatary:


heshamelmatary:
The rv32imac,gcc failure is likely due to old Binutils installed on the test machines. This could either be fixed by upgrading Binutils there, or maybe use "-z nostart-stop-gc" (as per the commit message) to the linker flags.

elfloader-tool/src/arch-arm/linker.lds
```diff
@@ -24,7 +18,6 @@ SECTIONS
         *(.text.start)
     }
 }
-INSERT BEFORE .text;
 
 SECTIONS
 {
```
kent-mcleod, line 23: What is the new section table layout after these changes?

This is what currently seems to be produced:
```
  0 .start        00000034  85000000  85000000  00010000  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  1 .text         000073c4  85001000  85001000  00011000  2**12
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  2 .rodata       00000ec7  850083c4  850083c4  000183c4  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  3 .ARM.exidx    00000008  8500928c  8500928c  0001928c  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
  4 .bss          00010048  8500c000  8500c000  00019294  2**14
                  ALLOC
  5 ._archive_cpio 01842a00  8501c048  8501c048  0002c048  2**0
                  CONTENTS, ALLOC, LOAD, DATA
  6 .data         0000002c  8686ea48  8686ea48  0187ea48  2**2
                  CONTENTS, ALLOC, LOAD, DATA
  7 _driver_list  00000030  8686ea74  8686ea74  0187ea74  2**2
                  CONTENTS, ALLOC, LOAD, DATA

```

elfloader-tool/src/arch-arm/linker.lds
```diff
@@ -28,7 +28,7 @@ INSERT BEFORE .text;
...
-    .bss : {
+    .bss (NOLOAD) : {
```
kent-mcleod, line 24: Can you add a note to this commit message that NOLOAD is already implicitly added by the current toolchain.

cmake-tool/helpers/tls_rootserver.lds
```diff
@@ -12,7 +12,7 @@ SECTIONS
         *(.tdata .tdata.* .gnu.linkonce.td.*)
         _tdata_end = . ;
     }
-    .tbss :
+    .tbss (NOLOAD):
```
kent-mcleod, line 15: This is already implicitly present in the current toolchain right? 
```
  5 .tbss         0000002c  016c3b90  016c3b90  016b3b8f  2**3
                  ALLOC, THREAD_LOCAL
```

kent-mcleod:
Err, I may be mixing up the meaning of `(NOLOAD)` in the linker script and the meaning of the `ALLOC` and `LOAD` flags in the section table, and what they mean about whether contents are allocated in the ELF file data, allocated in the program image and initialized by the loader (or zeroed).  And then what the implicit rules are with .bss, .tbss and other special named sections in the GCC toolchains.   Would you be able to add a summary @heshamelmatary ?

heshamelmatary:
> Err, I may be mixing up the meaning of `(NOLOAD)` in the linker script and the meaning of the `ALLOC` and `LOAD` flags in the section table, and what they mean about whether contents are allocated in the ELF file data, allocated in the program image and initialized by the loader (or zeroed). And then what the implicit rules are with .bss, .tbss and other special named sections in the GCC toolchains. Would you be able to add a summary @heshamelmatary ?

Most linkers heuristically recognise common section names and apply the proper flags to them. .[t]bss sections for example should be recognised and linkers mark them ALLOC, NOLOAD, which means that they don't occupy space in the ELF image and shouldn't be loaded at run-time, but they should be allocated space at run-time in the process image when loaded.

Adding NOLOAD (similar to the seL4's kernel) is good for intentionality, and for linkers that might not automatically flag .[t]bss at NOLOAD. I've come across a bug in lld that actually allocated space in the ELF before, but I believe that got fixed.

I will add that explanation to the commit message.

cmake-tool/helpers/tls_rootserver.lds
heshamelmatary: Yes that's correct for lld too. I just added NOLOAD here for intentionality and to avoid potential bugs with other linkers and/or older ones.

elfloader-tool/src/arch-arm/linker.lds
heshamelmatary: lld (unlike ld.bfd) does place .rodata before .text, the layout after the changes are:

`[ 1] .rodata.str1.1    PROGBITS        0000000040a11000 001000 000b45 01 AMS  0   0  1
  [ 2] .rodata           PROGBITS        0000000040a11b48 001b48 000668 00   A  0   0  8
  [ 3] .start            PROGBITS        0000000040a121b0 0021b0 00002c 00  AX  0   0  1
  [ 4] .text             PROGBITS        0000000040a13000 003000 007214 00  AX  0   0 4096
  [ 5] .bss              NOBITS          0000000040a1b000 00b000 006090 00  WA  0   0 4096
  [ 6] ._archive_cpio    PROGBITS        0000000040a21090 011090 486c00 00  WA  0   0  1
  [ 7] _driver_list      PROGBITS        0000000040ea7c90 497c90 000060 00 WAR  0   0  8
  [ 8] .data             PROGBITS        0000000040ea7cf0 497cf0 000150 00  WA  0   0  8`

heshamelmatary:
I dropped custom Arm's linker handling for sections and directives and rather added a new commit that unifies the linker script between Arm and RISC-V to include fixes for both and to make it more portable to build with lld and ld

heshamelmatary:
Also I modified the "retain" attribute to check if the toolchain supports it or not as this may fail on older GCC versions

elfloader-tool/src/arch-arm/linker.lds
kent-mcleod: The last thing I want to check on this is that I think there's a requirement that the .start section (with the entrypoint) is located first so that when the `ElfloaderImageBinary` option is set, the first address is the entry point.

kent-mcleod: So I checked on aarch64 with ld.ldd and the elfloader binary produced does have .rodata before the .start entry point. It's not a problem when the elfloader is loaded as an ELF and the bootloader can jump into the entry point. But for configurations that load it as a `ElfloaderImageBinary`, the entry point needs to be at the very start of the first loadable segment. We probably should just add a full linker script for arm like there is for riscv.

heshamelmatary: Hi @kent-mcleod 

This file no longer exists after I unified a full linker script for both Arm and RISC-V here https://github.com/seL4/seL4_tools/pull/173/commits/843b0c9d0f192aa9cf85fa7cdc27c5f8c4d0aa15

It does place `.rodata` after `.text.start` and `.text` as you expect. I do have a slight local change that I tested with all architectures and variants I have. I could push if you think the unified kernel linker is a good idea.

kent-mcleod: Sure. Currently it looks like only RISCV64 has .rodata after .text.start because of a #define around the .rodata declarations in the shared script.

heshamelmatary: @kent-mcleod Could you have a look at the latest commit? Yes, I preserved that #ifdef for RISC-V's `.rodata`

kent-mcleod: I have, and I set up a PR to check the change against the CI platforms where several fail to boot with the linker script changes: https://github.com/seL4/seL4/pull/1396

I think what's needed is to keep the scripts separate, but expand the arm script to be a full script rather than using the INSERT AFTER commands.

kent-mcleod: I'm planning on working out these issues over the next few days

kent-mcleod: I think rebasing this it now has no pending issues. I opened a separate PR to test: https://github.com/seL4/seL4_tools/pull/219 (and https://github.com/seL4/seL4/pull/1396)

kent-mcleod: A contribution of https://github.com/seL4/seL4_tools/pull/135 was to put the .rodata section back as an unconditional section which should be sufficient for _start remaining at the start.

kent-mcleod: This change works in the rebase branch now so I'm happy to merge either of them

kent-mcleod (approved):

kent-mcleod:
This has been now merged via: https://github.com/seL4/seL4_tools/pull/219.

I think now all of the lld patches have been merged :tada:

heshamelmatary:
Great, thanks, Kent!

PR #175, opened 2023-08-22
elfloader: Add ElfloaderPrecompile option

kent-mcleod:
This option generates an intermediate build file, elfloader.o, that can be combined in a separate build step with an archive.o containing the cpio archive with the kernel and rootserver images to load. This enables the elfloader to be installed by copying the elfloader.o and generated linker file to an install location and a future build step can link with a supplied archive.o into a bootable elfloader image using something like:
  aarch64-unknown-linux-gnu-gcc -march=armv8-a elfloader.o archive.o \
    -o elfloader -Wl,-T linker.lds_pp -nostdlib -static -lgcc

### DIFF
--- a/elfloader-tool/CMakeLists.txt
+++ b/elfloader-tool/CMakeLists.txt
@@ -95,6 +95,14 @@ config_option(
     DEPENDS KernelArchArmV8a
 )
 
+config_option(
+    ElfloaderPrecompile ELFLOADER_PRECOMPILE
+    "Precompile the elfloader but only partially link. Produces elfloader.o and a linker script.
+     Final linking can be done later with the CPIO archive."
+    DEFAULT OFF
+    DEPENDS "NOT ElfloaderImageEFI"
+)
+
 add_config_library(elfloader "${configure_string}")
 
 add_compile_options(-D_XOPEN_SOURCE=700 -ffreestanding -Wall -Werror -W -Wextra)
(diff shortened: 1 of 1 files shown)


elfloader-tool/CMakeLists.txt
```diff
@@ -383,6 +403,17 @@ if(ElfloaderImageEFI)
             # EFI_SUBSYSTEM=0xa indicates that we're building an EFI application.
             " -Wl,-T ${linkerScript} -nostdlib -shared -Wl,-Bsymbolic,--defsym=EFI_SUBSYSTEM=0xa -Wl,--build-id=none"
     )
+elseif(ElfloaderPrecompile)
+    add_custom_command(
+      OUTPUT elfloader.o
```
axel-h, line 408: Could you put `elfloader.o` in a CMake variable, so we don't have the name hard-coded everywhere?

kent-mcleod: What would the benefit be? We don't use a variable name for "kernel.elf", it's already an atom and the extra variable assignment is essentially redundant. The variable name would essentially just be `set(elfloader_o elfloader.o)` -> `${elfloader_o}` (`elfloader.o` is technically even a valid variable name).

mbrcknl (approved):
Looks good to me.

Though I wonder, would it make sense to avoid the config option, and always define separate targets for the library and the executable versions of the elfloader? (My knowledge of the CMake build system is limited, so it's fine if the answer is no.)

lsf37 (approved):
Looks good from my side (modulo style checker warnings).

axel-h:
> Looks good to me.
> 
> Though I wonder, would it make sense to avoid the config option, and always define separate targets for the library and the executable versions of the elfloader? (My knowledge of the CMake build system is limited, so it's fine if the answer is no.)

Why not even make the binary now use the *.o artifact? So, there would be separate lib and elf targets. Any outer build flow could just pick the target it wants then. Existing build flows would get the binary and see no difference. Assuming the intermediate step of building a lib and then the elf from that does not change the binary.

kent-mcleod:
> Though I wonder, would it make sense to avoid the config option, and always define separate targets for the library and the executable versions of the elfloader? (My knowledge of the CMake build system is limited, so it's fine if the answer is no.)

The config option means its easy to see that the refactor introduced by the new behavior doesn't change any existing behavior without the config option set.  Allowing both to exist at the same time would require rewriting more of the file to handle the need for both targets to have the same compilation target settings, but different linking target settings.

> Why not even make the binary now use the *.o artifact? So, there would be separate lib and elf targets. Any outer build flow could just pick the target it wants then. Existing build flows would get the binary and see no difference. Assuming the intermediate step of building a lib and then the elf from that does not change the binary.

Because I'm not 100% confident that they produce the same binary and didn't want to affect existing behavior.

PR #186, opened 2024-01-15
lld: Enable building C projects without cxx crt files

heshamelmatary:
crt[begin|end].o files are used for C++ projects. While they usually exist in GCC, they may or may not exist in LLVM builds that don't rely on GCC toolchain, and typically require a separate (optional) compiler-rt/cxx library to build.

This commit allows building projects without crt[begin|end].o files if they do not exist.

### DIFF
--- a/cmake-tool/helpers/environment_flags.cmake
+++ b/cmake-tool/helpers/environment_flags.cmake
@@ -92,6 +92,15 @@ macro(find_libgcc_files)
     # find the compilers crtbegin and crtend files
     gcc_print_file_name(CRTBeginFile crtbegin.o)
     gcc_print_file_name(CRTEndFile crtend.o)
+
+    # Empty crt files variables if not found in the current toolchain.
+    # This should allow building/linking without cxx crt files (e.g., just C projects).
+    if("${CRTBeginFile}" STREQUAL "crtbegin.o" OR "${CRTEndFile}" STREQUAL "crtend.o")
+        message(WARNING "cxx: no crt object files found.")
+        set(CRTBeginFile "")
+        set(CRTEndFile "")
+    endif()
+
     gcc_print_file_name(libgcc_eh libgcc_eh.a)
     set(libgcc "-lgcc")
     if(NOT "${libgcc_eh}" STREQUAL "libgcc_eh.a")


cmake-tool/helpers/environment_flags.cmake
```diff
@@ -92,6 +92,15 @@ macro(find_libgcc_files)
     # find the compilers crtbegin and crtend files
     gcc_print_file_name(CRTBeginFile crtbegin.o)
     gcc_print_file_name(CRTEndFile crtend.o)
+
+    # Empty crt files variables if not found in the current toolchain.
+    # This should allow building/linking without cxx crt files (e.g., just C projects).
+    if("${CRTBeginFile}" STREQUAL "crtbegin.o" OR ${CRTEndFile} STREQUAL "crtend.o")
```
axel-h, line 98: nitpicking: use quotes consistently
```suggestion
    if("${CRTBeginFile}" STREQUAL "crtbegin.o" OR "${CRTEndFile}" STREQUAL "crtend.o")
```

cmake-tool/helpers/environment_flags.cmake
```diff
@@ -92,6 +92,15 @@ macro(find_libgcc_files)
     # find the compilers crtbegin and crtend files
     gcc_print_file_name(CRTBeginFile crtbegin.o)
     gcc_print_file_name(CRTEndFile crtend.o)
+
+    # Empty crt files variables if not found in the current toolchain.
+    # This should allow building/linking without cxx crt files (e.g., just C projects).
```
axel-h, line 97: Can it happen that just one file of them exists? In this case we also drop it, should we keep it then?

heshamelmatary: I don't think so, they usually have to be bundled together for instructors/destructors

axel-h: The docs say it output the absolute name of the file. It does not say what we get when the file does not exist. Looking at `gcc_print_file_name()` is seem it prints the plain input when the file is not found, but does not check the return code actually.. Have you checked if we we get an error return code? Then we could have `gcc_print_file_name()` log a more detailed warning message and set the variable to empty. That seems be a bit cleaner than doing it here. It might have some side effects on the other places where this is used, but I assume all they run into the same issues basically if the input files does not exists so making the variable empty seem a sane choice.

heshamelmatary: > The docs say it output the absolute name of the file. It does not say what we get when the file does not exist. Looking at `gcc_print_file_name()` is seem it prints the plain input when the file is not found, but does not check the return code actually.. Have you checked if we we get an error return code? Then we could have `gcc_print_file_name()` log a more detailed warning message and set the variable to empty. That seems be a bit cleaner than doing it here. It might have some side effects on the other places where this is used, but I assume all they run into the same issues basically if the input files does not exists so making the variable empty seem a sane choice.

Yeah it just prints `crtend.o` or `crtbegin.o` if the file doesn't exist, with a success (0) return code, unfortunately

axel-h: > Yeah it just prints `crtend.o` or `crtbegin.o` if the file doesn't exist, with a success (0) return code, unfortunately

So, should we still try to handle things within `gcc_print_file_name()` then and make the variable empty instead of returning the input value. It would be a bigger change, but it make the general behavior more sane then for the caller.

heshamelmatary: I think it's the caller responsibility to handle such case in line with GCC/LLVM behaviour, as with anyone calling GCC/LLVM with `--print-file-name` argument, otherwise `--print-file-name` would have returned an error code in the first place. For instance, if the caller is a C++ project and must expect these files, it should fail the build, rather than set these files empty

axel-h (approved):
Looks ok to me.

heshamelmatary:
> crtbegin is also used in C for code marked as init code, although that is more rare. Won't this change break all such code marked with `__attribute__ ((constructor))`?

It shouldn't break any C code marked as constructor that sel4[runtime] might be already handling.

crt[begin|end] should be OS-specific and used in C++ or when dynamically loading C or C++ shared objects. seL4 doesn't (fully) support either AFAIK. The fact that the current build system is using GCC's is assuming seL4 userspace will imitate Linux as seL4 usually uses *-linux-* toolchain. IMO, if seL4 userspace wants to support proper C++ or dynamic loading of shared objects in the future, it should define its own crt[begin|end] files, not using whatever the toolchain provides.

Indanz:
> > crtbegin is also used in C for code marked as init code, although that is more rare. Won't this change break all such code marked with `__attribute__ ((constructor))`?
> 
> It shouldn't break any C code marked as constructor that sel4[runtime] might be already handling.
> 
> crt[begin|end] should be OS-specific and used in C++ or when dynamically loading C or C++ shared objects.

That seems a wrong assumption to me.

> seL4 doesn't (fully) support either AFAIK. The fact that the current build system is using GCC's is assuming seL4 userspace will imitate Linux as seL4 usually uses _-linux-_ toolchain. IMO, if seL4 userspace wants to support proper C++ or dynamic loading of shared objects in the future, it should define its own crt[begin|end] files, not using whatever the toolchain provides.

Agreed, but isn't this file also used for the root server environment, which may use the default crtbegin files? And in that case constructor/destructor attributes may work now, but not after this change?

heshamelmatary:
> > > crtbegin is also used in C for code marked as init code, although that is more rare. Won't this change break all such code marked with `__attribute__ ((constructor))`?
> > 
> > 
> > It shouldn't break any C code marked as constructor that sel4[runtime] might be already handling.
> > crt[begin|end] should be OS-specific and used in C++ or when dynamically loading C or C++ shared objects.
> 
> That seems a wrong assumption to me.
> 
> > seL4 doesn't (fully) support either AFAIK. The fact that the current build system is using GCC's is assuming seL4 userspace will imitate Linux as seL4 usually uses _-linux-_ toolchain. IMO, if seL4 userspace wants to support proper C++ or dynamic loading of shared objects in the future, it should define its own crt[begin|end] files, not using whatever the toolchain provides.
> 
> Agreed, but isn't this file also used for the root server environment, which may use the default crtbegin files? And in that case constructor/destructor attributes may work now, but not after this change?

No, this file isn't used for the root task env; the root task has its own [se4_]crt[i].S files. sel4test and sel4bench build and pass fine with that change,

Indanz (approved):

PR #190, opened 2024-02-08
Elfloader: NVIDIA Jetson Orin support

andybui01:
Elfloader portion of the Jetson Orin port effort (https://github.com/seL4/seL4/pull/1135)

This patch set is quite big and might be hard to review, if the reviewers would like me to break this into separate PRs please let me know.

### DIFF
--- a/elfloader-tool/include/arch-arm/elfloader.h
+++ b/elfloader-tool/include/arch-arm/elfloader.h
@@ -22,7 +22,22 @@ typedef void (*init_arm_kernel_t)(word_t ui_p_reg_start,
 
 /* Enable the mmu. */
 extern void arm_enable_mmu(void);
+
+/* These functions are very similar however, there are some small differences
+ * between the ARMv8 and legacy implementation.
+ *
+ * New ARMv8 implementation:
+ * - Does the MMU disabling. This is to keep the time spent with MMU off low.
+ * - Is only meant if seL4 runs in EL2.
+ */
+#if defined(CONFIG_ARCH_ARM_V8A)
+/* Switches MMU-related stuff: pagetables, MAIR & TCR etc. Works also if the MMU
+ * was off initially. EL2 translation regime only.
+ */
+extern void arm_switch_to_hyp_tables(void);
+#else
 extern void arm_enable_hyp_mmu(void);
+#endif
 
 
 /* Setup boot VSpace. */
--- a/elfloader-tool/src/arch-arm/drivers/smp-psci.c
+++ b/elfloader-tool/src/arch-arm/drivers/smp-psci.c
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
+#include <autoconf.h>
 #include <elfloader_common.h>
 #include <devices_gen.h>
 #include <drivers/common.h>
@@ -24,7 +25,13 @@ static int smp_psci_cpu_on(UNUSED struct elfloader_device *dev,
     }
     secondary_data.entry = entry;
     secondary_data.stack = stack;
-    dmb();
+#if defined(CONFIG_ARCH_AARCH64)
+    /* If the secondary core caches are off, need to make sure that the info
+     * is clean to the physical memory so that the sedcondary cores can read it.
+     */
+    asm volatile("dc cvac, %0" :: "r"(&secondary_data));
+    dsb();
+#endif
     int ret = psci_cpu_on(cpu->cpu_id, (unsigned long)&secondary_startup, 0);
     if (ret != PSCI_SUCCESS) {
         printf("Failed to bring up core 0x%x with error %d\n", cpu->cpu_id, ret);
--- a/elfloader-tool/src/arch-arm/smp_boot.c
+++ b/elfloader-tool/src/arch-arm/smp_boot.c
@@ -35,7 +35,11 @@ void non_boot_main(void)
 #endif
     /* Spin until the first CPU has finished initialisation. */
     while (!non_boot_lock) {
-#ifndef CONFIG_ARCH_AARCH64
+#ifdef CONFIG_ARCH_AARCH64
+        /* The compiler may optimize this loop away, add a dsb()
+         * to force a reload. */
+        dsb();
+#else
         cpu_idle();
 #endif
     }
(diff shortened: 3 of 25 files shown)


andybui01:
Ok, failing on all the ARMv8 32 bit builds, the change in `sys_boot.c` needs to be simplified since _really_ all we changed was ARMv8 64 bit code.

andybui01:
Two failures on the HW run:
- TX2 clang (all configs): cannot connect to mq?
- TQMA, passes all tests but output is cut short, CI thinks it failed?

Besides that, there are positive signs that the elfloader changes (and especially the aarch64 changes) don't introduce any regressions.

cc: @kent-mcleod @axel-h @Indanz I'd appreciate your reviews, let me know if the diff can be minimized or cleaned up

kent-mcleod:
> I'd appreciate your reviews, let me know if the diff can be minimized or cleaned up

I'd find it easier to fit in time and attention for reviewing if the PR was split up and merged incrementally. 

- The EFI commits and many of the bugfix commits that are self-contained fixes can likely be reviewed and merged with limited discussion required through their own PR.  Specifically:
  - 0108b774b0511bc81e4ccb86aae69aba36468aea
  - 7b3c2f32491a3491817e2196f4fa40d4f0c93284
  - 1aa2fb03cae603fdb3a83fb791fbcabe2fb866f5
  - c329bef78ba38a83aa0b97663fae96b18ce25601
  - e8a15ec23cf48b473133069a33674c15d2fc4d74
  - dea692439d0e8e3f07edf74ab63a845f11fcfee8
  - 65b49ae06483a221b269b622bb3d9c748b7890d6
  - fa81fff7f15e341737554d55f3731b9b88f31fe3
  - 33f00c8babeb5dd3014115cd8b6e5c5ac39a557b
  - 5ebb1fcef3b1a3813063336188a19dbb63d877c6
  - 80b8902c968523183e62eb6f90a0450a4e8c040a
  - 0e0a99d2a78880cd4afb7c73ee8ee81af1b2a767
- The Orin platform port commit (9f2a8d027d79777fa843e29dd5ee4e2f4f09821d) is trivial once all the dependent changes have been landed. This can probably be merged in advance of the other changes.
- The remaining commits will probably require the bulk of discussion in my opinion as they change structure of common code.

So if you are able to split out the above commits into a separate PR and leave the rest in this one I think that will help us proceed.

elfloader-tool/src/arch-arm/drivers/smp-psci.c
```diff
@@ -24,7 +25,13 @@ static int smp_psci_cpu_on(UNUSED struct elfloader_device *dev,
     }
     secondary_data.entry = entry;
     secondary_data.stack = stack;
-    dmb();
+#if defined(CONFIG_ARCH_AARCH64)
+    /* If the secondary core caches are off, need to make sure that the info
+     * is clean to the physical memory so that the sedcondary cores can read it.
```
Indanz, line 30: Typo: "sedcondary".

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -34,7 +34,11 @@ void non_boot_main(void)
 #endif
     /* Spin until the first CPU has finished initialisation. */
     while (!non_boot_lock) {
-#ifndef CONFIG_ARCH_AARCH64
+#ifdef CONFIG_ARCH_AARCH64
+        /* The compiler may optimize this loop away, add a dsb()
+         * to force a reload. */
+        dsb();
```
Indanz, line 40: This also applies to non-aarch64, so this doesn't look like the right solution. Using standard atomics may be better and more portable here.

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -141,6 +151,13 @@ void smp_boot(void)
     arm_disable_dcaches();
 #endif
     init_cpus();
+#if defined(CONFIG_ARCH_AARCH64)
+    dsb();
+    non_boot_lock = 1;
+    /* Secondary CPUs may still run with MMU & caches off. Force the update to be visible. */
```
Indanz, line 154: The same is true for 32-bit, right?

elfloader-tool/src/binaries/efi/gnuefi/elf_aarch64_efi.lds
```diff
@@ -61,6 +61,7 @@ SECTIONS
   .dynstr   : { *(.dynstr) }
   . = ALIGN(4096);
   .note.gnu.build-id : { *(.note.gnu.build-id) }
+  _end = .;
```
Indanz, line 67: This should be part of commit dea692439d0e8e3f07edf74ab63a845f11fcfee8 I think.

elfloader-tool/include/arch-arm/elfloader.h
```diff
@@ -22,7 +22,22 @@ typedef void (*init_arm_kernel_t)(word_t ui_p_reg_start,
...
+ *
+ * New ARMv8 implementation:
+ * - Does the MMU disabling. This is to keep the time spent with MMU off low.
+ * - Is only meant if seL4 runs in EL2.
+ */
+#if defined(CONFIG_ARCH_AARCH64)
+/* Switches MMU-related stuff: pagetables, MAIR & TCR etc. Works also if the MMU
+ * was off initially. EL2 translation regime only.
+ */
+extern void arm_switch_to_hyp_tables(void);
```
Indanz, line 37: Perhaps call it `aarch64_switch_to_hyp_tables` if there is no 32-bit version of it.

Indanz:
I agree with Kent.

I think 4996de7d5e086f894fa49d2f2b6e35a7d9c44aee is the most controversial change and hardest to review for correctness, all other commits seem fairly straightforward.

andybui01:
No problem, thanks guys.

What I'll do is split this into 3 PRs then:
1. General stability/EFI changes
2. The series of commits that changes ARM booting and splits aarch64 out, including the humongous page table commit (https://github.com/seL4/seL4_tools/commit/4996de7d5e086f894fa49d2f2b6e35a7d9c44aee)
3. Finally, the Jetson Orin port

Indanz:
I would prefer the problematic commit to be in its own PR, so we can discuss it there without holding up the rest.

PR #191, opened 2024-02-28
elfloader: improve stability

andybui01:
As mentioned in https://github.com/seL4/seL4_tools/pull/190, this is the series of commits that do general touch-ups in the elfloader.


elfloader-tool/src/arch-arm/drivers/smp-psci.c
```diff
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: GPL-2.0-only
  */
+#include <autoconf.h>
```
axel-h, line 6: We should include `autoconf.h` in `elfloader_common.h`, then we don't have to do this in every file explicitly.

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -134,6 +144,13 @@ void smp_boot(void)
     arm_disable_dcaches();
 #endif
     init_cpus();
+#if defined(CONFIG_ARCH_AARCH64)
```
axel-h, line 147: Could avoid the code duplication:

```
#if defined(CONFIG_ARCH_AARCH64)
    dsb();
#endif

    non_boot_lock = 1;

#if defined(CONFIG_ARCH_AARCH64)
    /* Secondary CPUs may still run with MMU & caches off. Force the update to be visible. */
    asm volatile("dc civac, %0\n\t" :: "r"(&non_boot_lock) : "memory");;
#endif
```

elfloader-tool/src/arch-arm/sys_boot.c
```diff
@@ -23,7 +23,7 @@
 #define DTB_MAGIC (0xedfe0dd0)
 
 /* Maximum alignment we need to preserve when relocating (64K) */
-#define MAX_ALIGN_BITS (14)
+#define MAX_ALIGN_BITS (16)
```
axel-h, line 26: It would be helpful  to have this parts of the commit comment as comment in the sources as a reference
/* The 64 kiB alignment is a maximum requirement for a stage2
 * concatenated pagetable. See Table G5-4 in ARM DDI 0487I.a, page
 * G5-9186.
 */

elfloader-tool/src/arch-arm/64/mmu.c
```diff
@@ -11,6 +11,8 @@
 #include <mode/structures.h>
 #include <printf.h>
 #include <abort.h>
+#include <mode/aarch64.h>
+#include <armv/machine.h>
```
axel-h, line 15: Why do we include this file now also, was it missing before?

elfloader-tool/include/arch-arm/64/mode/structures.h
```diff
@@ -21,9 +21,9 @@
 #define PMD_BITS                9
 #define PMD_SIZE_BITS           (PMD_BITS + PMDE_SIZE_BITS)
 
-#define GET_PGD_INDEX(x)        (((x) >> (ARM_2MB_BLOCK_BITS + PMD_BITS + PUD_BITS)) & MASK(PGD_BITS))
-#define GET_PUD_INDEX(x)        (((x) >> (ARM_2MB_BLOCK_BITS + PMD_BITS)) & MASK(PUD_BITS))
-#define GET_PMD_INDEX(x)        (((x) >> (ARM_2MB_BLOCK_BITS)) & MASK(PMD_BITS))
+#define GET_PGD_INDEX(x)        (((word_t)(x) >> (ARM_2MB_BLOCK_BITS + PMD_BITS + PUD_BITS)) & MASK(PGD_BITS))
```
axel-h, line 24: In the commit comment it says that this works for pointer now also, but pointer must be cast to `uintptr_t` first to make then an integer. Maybe add this here also then?

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -34,7 +34,11 @@ void non_boot_main(void)
 #endif
     /* Spin until the first CPU has finished initialisation. */
     while (!non_boot_lock) {
-#ifndef CONFIG_ARCH_AARCH64
+#ifdef CONFIG_ARCH_AARCH64
+        /* The compiler may optimize this loop away, add a dsb()
```
axel-h, line 38: The comment seems incorrect, the variable is declared `volatile`, so the compiler is not allowed to optimize this out. However, the barrier might sill be needed to force a OO-pipeline not to try smart things. Instead of the barrier, we might also consider using https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html to let the compiler pick the appropriate meachnism.

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -117,7 +121,13 @@ WEAK void init_cpus(void)
             abort();
         }
 
-        while (!is_core_up(num_cpus));
+        while (!is_core_up(num_cpus)) {
+#if defined(CONFIG_ARCH_AARCH64)
+            /* The compiler may optimize this loop away, add a dsb()
```
axel-h, line 126: Same here, see https://github.com/seL4/seL4_tools/pull/191/commits/31daf6ab48c8f837bbf772d8c5de354b6ced3e0a#r1506631069

elfloader-tool/include/arch-arm/cpuid.h
```diff
@@ -12,7 +12,7 @@
 uint32_t read_cpuid_id(void);
 
 /* read MP ID register from CPUID */
-uint32_t read_cpuid_mpidr(void);
```
axel-h, line 15: I wonder if we should drop the `cpuid.c` file completely and move all functions as `static inline` here. Seem they are just wrappers around some assembly instructions anyway.

axel-h:
The simulation on ARMv8 fails now, could you have a look at this? Maybe you should split this PR even further up, so we can merge the trivial things faster.

elfloader-tool/src/arch-arm/64/mmu.c
andybui01: I believe it's for some of the constants such as `MT_DEVICE_nGnRnE` or this change is just an artifact of splitting up the previous PR.

andybui01:
> The simulation on ARMv8 fails now, could you have a look at this?

The failure is due to the following reason: as it stands, the elfloader creates the boot page table by mapping 2 regions:
1. The kernel region (`kernel_info->virt_region_start`) as `MT_NORMAL` memory.
2. A bulk 512 GiB 1:1 mapping (used by the elfloader) that is strongly-ordered.

The final commit (https://github.com/seL4/seL4_tools/pull/191/commits/3166422ebc96ecc068df8ca854da589e31fd567c) erroneously assumes that this is region is for device memory, that is not correct. Instead, this commit will be part of the page table PR to follow, which splits the boot page table into distinct regions (elfloader - normal, kernel - normal, uart mmio - strongly ordered).

elfloader-tool/include/arch-arm/64/mode/structures.h
andybui01: I believe the reason behind the comment is that in `tools/seL4/elfloader-tool/include/types.h` "word_t is practically an alias for size_t/uintptr_t on the platforms we support so far". Do you mean to just change `word_t` to `uintptr_t` here?

kent-mcleod:
> As mentioned in #190, this is the series of commits that do general touch-ups in the elfloader.

This PR seems to not contain one of the touch up commits: https://github.com/seL4/seL4_tools/commit/0108b774b0511bc81e4ccb86aae69aba36468aea

And contains an additional commit that is not a touch up commit and changes behavior: https://github.com/seL4/seL4_tools/pull/191/commits/7e410f094417edcbbcb25b9148ae44cb21f70a63

Is this intentional?

andybui01:
> > As mentioned in #190, this is the series of commits that do general touch-ups in the elfloader.
> 
> This PR seems to not contain one of the touch up commits: [0108b77](https://github.com/seL4/seL4_tools/commit/0108b774b0511bc81e4ccb86aae69aba36468aea)
> 
> And contains an additional commit that is not a touch up commit and changes behavior: [7e410f0](https://github.com/seL4/seL4_tools/commit/7e410f094417edcbbcb25b9148ae44cb21f70a63)
> 
> Is this intentional?

Yeah, the first one makes less sense to me without the other commits that all attempt to solve a problem (stabilize aarch64 booting for the Orin).

As for the second, I think its impact is low enough to be included here? It doesn't seem to be that big of a behavioral change.

elfloader-tool/src/arch-arm/smp_boot.c
andybui01: Yes you're correct it's declared `volatile` so this is not needed. I can add a `asm volatile("": : :"memory")` to the loop body to let the compiler handle this?

elfloader-tool/include/arch-arm/cpuid.h
andybui01: I think this can be in a general cleanup PR, same with the `autoconf.h` change

elfloader-tool/src/binaries/efi/gnuefi/elf_aarch64_efi.lds
```diff
@@ -31,6 +31,9 @@ SECTIONS
    *(.data)
    *(.data1)
    *(.data.*)
+   __start__driver_list = .;
+   *(_driver_list)
```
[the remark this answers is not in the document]
andybui01: This is within the riscv linker script:
```
    . = ALIGN(16);
    .rodata :
    {
        *(.srodata*)
        . = ALIGN(16);
        *(.rodata)
        *(.rodata.*)
        /*
         * ld crashes when we add this here: *(_driver_list)
         */
        . = ALIGN(16);
        _archive_start = .;
        *(._archive_cpio)
        _archive_end = .;
    }
```

Moving the `_driver_list` into rodata may be breaking.

elfloader-tool/include/arch-arm/64/mode/structures.h
axel-h: I think strict C rules say, that pointers must be cast to `uintptr_t` to make them an integer, from there you can cast the integer to any other integer type.

elfloader-tool/src/arch-arm/smp_boot.c
axel-h: Please fix the comment here also. I think at this point we should use the same cache settings on all core at least, so the dsb is really enough. And again, removing the `#if defined(CONFIG_ARCH_AARCH64)` and doing this on ARmv7 as well can't hurt.

elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -134,6 +144,17 @@ void smp_boot(void)
...
     init_cpus();
+
+#if defined(CONFIG_ARCH_AARCH64)
+    dsb();
+#endif
+
     non_boot_lock = 1;
+
+#if defined(CONFIG_ARCH_AARCH64)
+    /* Secondary CPUs may still run with MMU & caches off. Force the update to be visible. */
```
axel-h, line 155: The comment applies in general, not just for CONFIG_ARCH_AARCH64. Seems just for AARCH64 we have the proper command here, for ARMv7/AARCH there is still a open ToDo to look up the proper command/function one day can call it here.

elfloader-tool/src/binaries/efi/gnuefi/elf_aarch64_efi.lds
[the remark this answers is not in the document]
axel-h: Code the commit history give something why ld is crashing here? Maybe this is fixed by now?

elfloader-tool/src/arch-arm/smp_boot.c
andybui01: Sure, I'll move the comment out

elfloader-tool/src/arch-arm/sys_boot.c
```diff
@@ -216,17 +221,23 @@ void continue_boot(int was_relocated)
...
 
-    /* Enter kernel. The UART may no longer be accessible here. */
+    /* The UART may no longer be accessible here. */
     if ((uintptr_t)uart_get_mmio() < kernel_info.virt_region_start) {
         printf("Jumping to kernel-image entry point...\n\n");
     }
 
+#if defined(CONFIG_ARCH_AARCH64)
+    /* Clear D&A in DAIF */
+    asm volatile("msr daifclr, #0xC\n\t");
```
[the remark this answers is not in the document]
andybui01: I'll move this into something like `arm_exit_hook()` then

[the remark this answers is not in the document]
axel-h: Actually I just meant moving the the `asm volatile("msr daifclr, #0xC\n\t")` into some wrapper. I don't think there is much gain making an `arm_exit_hook()` that is called here, as the only place where we call that is here anyway?

elfloader-tool/include/arch-arm/64/mode/structures.h
andybui01: That is reasonable, I did find this excerpt in the C99 standard though, and it's the only part I could find on `uintptr_t`:

> The following type designates an unsigned integer type with the property that any valid
pointer to void can be converted to this type, then converted back to pointer to void,
and the result will compare equal to the original pointer: uintptr_t

So the strictest form would be a cast to `void *` first before `uintptr_t`. However, I think all the reasonable implementations have sensible ways for any pointer casted to `uintptr_t` to behave such that we can skip the `void *` cast.

elfloader-tool/src/arch-arm/smp_boot.c
andybui01: cpu_idle() is just `dsb()` and `wfi()` paired, so we can make this common.

axel-h: I don't understand. `cpu_idle()` is doing something completely different, so this should not be used here.

andybui01: The current code as is
```
    while (!non_boot_lock) {
#ifdef CONFIG_ARCH_AARCH64
        dsb();
#else
        cpu_idle();
#endif
    }
```

And since `cpu_idle()` is itself: dsb() + wfi(), the dsb() would already be common.

axel-h: Ah ok, I see. But It's a bit odd that ARMv7 is doing a `wfi()` here also - I can't see who triggers the event on the other core that will bring us out of the `wfi` then. Did this ever work or was this stuck forever then - or did something else trigger an even by chance? I can't find anything in the commit history that explains this.

andybui01: Maybe this can be simplified to just a barrier and we can find a way to do a cache invalidation (not just aarch64) on `non_boot_lock` when the boot CPU finishes initializing?

elfloader-tool/src/arch-arm/smp_boot.c
axel-h: I wonder why you do the `dsb()` before updating `non_boot_lock` actually.  This should get a comment that explains this. Seems we want to ensure all changes done far get visible, before we release the other cores. We should do the `dsb()` unconditionally then, not just on aarch64.

```
    /* ensure all operation complete before we release the other cores */
    dsb(); 
    /* release secondary cores */
    non_boot_lock = 1;
    /* Secondary cores may still run with MMU & caches off. Force the update to be visible. */ 
#if defined(CONFIG_ARCH_AARCH64) 
    asm volatile("dc civac, %0\n\t" :: "r"(&non_boot_lock) : "memory"); 
#else
    // ToDo: add flush operation here to be safe.
#endif

elfloader-tool/src/arch-arm/smp_boot.c
axel-h: Since on aarch64 the D-Cache is disabled before the loop anyway, just a `dsb()` should be sufficient here. And it seems this worked well for your colleagues when the made the code change. 
Since I'm not really sure how this is supposed to work on ARMv7/AARCH32 and we can't test this now either, it might be best to just leave it and add a comment that calling `cpu_idle()` this looks like a hack that should either be explained or reworked.

elfloader-tool/src/binaries/efi/gnuefi/elf_aarch64_efi.lds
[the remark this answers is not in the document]
andybui01: Commit history didn't give any hints unfortunately. I don't have the risc-v toolchain currently so I'll try and build when I get the chance.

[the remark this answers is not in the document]
axel-h: I will also have a look on RISC-V.

PR #194, opened 2024-03-14
elfloader/riscv: always pass hart/core information

axel-h:
Align the SMP and non-SMP ABI for kernel entry and pass on the hard ID obtained from the previous bootloader unconditionally. This allows the kernel to use the ID, e.g. when setting up the PLIC, so there is no need to require defining a platform specific hard-codes value. Furthermore, this change allows running a single core kernel in SMP configuration also.

The long term goal if this change is removing the need to have the explicit `CONFIG_FIRST_HART_ID` in the seL4 kernel. It will  just use the ID that is passed (unless platforms configure something else due to special needs). That aligns the behavior a bit to ARM, where we can read the current core ID from a register in case it is needed somewhere, so there is no need to configure this explicitly. Unfortunately, RISC-V does not have such a register here.


axel-h:
Is there a chance to get this merged?

axel-h:
> Looks good to me, however, care must be taken before this is merged, as it requires an ABI change on the seL4 kernel side too.

This just adds parameters that have been undefined before. So merging this is not expected to break anything. It's just when we start using this parameters in the next boot stage, then there will be a dependency on the new ABI. 

So, are you going to merge it?

Indanz:
> This just adds parameters that have been undefined before. So merging this is not expected to break anything. It's just when we start using this parameters in the next boot stage, then there will be a dependency on the new ABI.

Yes, you're right. For some reason I thought it changed existing ABI too.

PR #195, opened 2024-03-22
elfloader: use loop variable as index

axel-h:
Using the loop variable here seem much easier. This is extracted from a larger patch set, removing usage of `num_images`  simplifies further refactoring.

### DIFF
--- a/elfloader-tool/src/common.c
+++ b/elfloader-tool/src/common.c
@@ -599,7 +599,7 @@ int load_images(
                        "app.bin", // hash file
                        next_phys_addr,
                        1,  // keep ELF headers
-                       &user_info[*num_images],
+                       &user_info[i],
                        &next_phys_addr);
         if (0 != ret) {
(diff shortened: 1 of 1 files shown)


Indanz:
I disagree, this change on its own is not enough of a cleanup to justify all the hassle (PR+review+running CI test etc.), there is way too much overhead.

The number of images loaded is only used for:

```
if (num_apps != 1) {
        printf("ERROR: expected to load just 1 app, actually loaded %u apps\n",
               num_apps);
        return -1;
    }
```
Which is moronic, if we want to only load one user image, then just do that.

Now if there is a real error during loading a user image, it doesn't get returned to the caller. The whole API is silly. So there is a lot of room for real improvement in this code, both to make it clearer and simpler. We don't need to do that of course, but if you want to improve it, please make a significant change.

I'm going to spend my time on letting seL4 kernel boot itself instead though.

PR #197, opened 2024-03-22
elfloader: make check more intuitive

axel-h:
This has been extracted from a bigger patchset to improve the code maintainability/extendability.

### DIFF
--- a/elfloader-tool/src/arch-arm/smp_boot.c
+++ b/elfloader-tool/src/arch-arm/smp_boot.c
@@ -40,8 +40,9 @@ void non_boot_main(void)
     }
 
     /* Do any driver specific non_boot core init */
-    if (initialise_devices_non_boot()) {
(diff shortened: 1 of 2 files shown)


elfloader-tool/src/arch-arm/smp_boot.c
```diff
@@ -40,7 +40,7 @@ void non_boot_main(void)
     }
 
     /* Do any driver specific non_boot core init */
-    if (initialise_devices_non_boot()) {
+    if (0 != initialise_devices_non_boot()) {
```
lsf37, line 43: I don't find this more intuitive. The return type of `initialise_devices_non_boot` is currently `int`, but it is only used as bool. Should we not change it to bool instead?

axel-h: You have a point here. The current scheme allows returning different error codes, I can make this even more clear and then also print the code in the message.

PR #198, opened 2024-04-09
elfloader: initialise stack allocated struct

Ivan-Velickovic:
Only results in a compilation error on newer GCC versions.

What confuses me though is this only happens for me in sel4bench, but not sel4test. I believe I'm using the same compiler in both projects, but se4lbench results in a unitialised warning that is treated as a compile error. sel4test does not even print the warning?

### DIFF
--- a/elfloader-tool/src/plat/tx2/platform_init.c
+++ b/elfloader-tool/src/plat/tx2/platform_init.c
@@ -61,7 +61,7 @@ static __attribute__((noinline)) int send_smc(uint8_t func, struct mce_regs *reg
 
 static void tegra_mce_write_uncore_mca(mca_cmd_t cmd, uint64_t data, uint32_t *err)
 {
-    struct mce_regs regs;
+    struct mce_regs regs = {0};
     regs.args[0] = cmd.data;
     regs.args[1] = data;
     send_smc(13, &regs);


elfloader-tool/src/plat/tx2/platform_init.c
```diff
@@ -61,7 +61,7 @@ static __attribute__((noinline)) int send_smc(uint8_t func, struct mce_regs *reg
 
 static void tegra_mce_write_uncore_mca(mca_cmd_t cmd, uint64_t data, uint32_t *err)
 {
-    struct mce_regs regs;
+    struct mce_regs regs = {0};
```
axel-h, line 64: If gcc complain here, it should also complain about line 74 below in `enable_serr()`, where `mca_cmd_t cmd` has uninitialized fields, as it is also a stack variable that will start with content and we don't set all members. We should initialize this will all zeros also.

lsf37: gcc might be smart enough to figure out that only the initialised fields are used from the one in line 74, esp if it inlines it. In fact, it appears `enable_serr` sets two fields of that struct/union that are never used. `regs` here on the other hand is passed to an `asm` block, so gcc is unlikely to know more.

lsf37 (approved):

PR #199, opened 2024-04-15
cmake: use quotes

axel-h:
Avoid odd error messages about an invalid number of parameters in case `TRIPLE` is empty. This happens with a customized build system.


lsf37 (approved):
Looks good from my side, just waiting for the broader build test to pass.

PR #200, opened 2024-04-26
misc: remove deprecated astyle option

Ivan-Velickovic:
Release 3.0 of astyle replaced this option with
'max-continuation-indent'. The replaced option is already in the astylerc so we just need to remove the option here.

Further information can be found in the release notes: https://astyle.sourceforge.net/news.html

### DIFF
--- a/misc/astylerc
+++ b/misc/astylerc
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD-2-Clause
 #
 
-max-instatement-indent=120
 style=otbs
 pad-header
 indent=spaces=4


lsf37 (approved):
Happy with removing it, but it means we need to update the recommended version in the style guide (which is ancient). It looks like the output is stable, so I'll put up a PR for that.

lsf37:
Just for posterity: the issue is not that the option has been replaced in 3.0, it is that it is now no longer recognised in more recent version. We had both for a while, but now only `astyle` versions >= 3.0 will work.

PR #202, opened 2024-06-25
Enable SMP on qemu-arm-virt

andybui01:
See: https://github.com/seL4/seL4/pull/1275


cmake-tool/helpers/simulation.cmake
```diff
@@ -144,14 +144,19 @@ function(GenerateSimulateScript)
         set(qemu_sim_extra_args "-bios none")
     elseif(KernelPlatformQEMUArmVirt)
         set(QemuBinaryMachine "qemu-system-${QEMU_ARCH}")
-        if(KernelArmHypervisorSupport)
+        if(Kernel32)
             set(sim_machine "virt,virtualization=on,highmem=off,secure=off")
         else()
-            set(sim_machine "virt")
+            set(sim_machine "virt,virtualization=on,highmem=on,secure=off")
```
lsf37, line 150: Similar to the seL4 PR, I don't quite understand why would always switch on virtualisation. We do want to test simulated kernels on platforms that don't support virtualisation, so I don't think we should do that.

Indanz:
Feel free to re-open if adapting Elfloader doesn't work.

PR #204, opened 2024-07-15
misc: check that astyle exists

Ivan-Velickovic:
We use the seL4 style check for sDDF and most people who try to run the style check offline end up running into weird issues since they don't have it installed.

### DIFF
--- a/misc/style-c.sh
+++ b/misc/style-c.sh
@@ -4,6 +4,13 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 #
+set -e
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
👍

`astyle` is of course described as a dependency in the setup instructions, but let's not rely on that.

Ivan-Velickovic:
> astyle is of course described as a dependency in the setup instructions, but let's not rely on that.

I have it in the sDDF CI README as well but it's not enough :)

PR #208, opened 2024-10-08
simulation: Add support for i.mx7 simulation through Qemu

jcdubois:
Qemu now supports i.mx7d Sabre platform.


lsf37:
Thanks for that. When you say "now", which version of qemu do you mean? Can you pinpoint from which version on support was added?

jcdubois:
Initial i.MX7 support is quite "old" as it started in 2018 in Qemu but I believe it has a reasonable support since at least august 2023.

Anyway I was able to run successfully seL4test on emulated i.MX7 Sabre PTF with the latest Qemu package from Ubuntu (version 6.2.0 of Qemu).

kent-mcleod (approved):

PR #213, opened 2025-02-03
Support for Cheshire (CVA6)

omeh-a:
Prevents building OpenSBI automatically in Cmake builds.


cmake-tool/helpers/rootserver.cmake
Ivan-Velickovic, line 1: We don't need this change I think.

omeh-a: Just tested, the change in application_settings indeed does the trick. Removed.

Ivan-Velickovic: It's still here.

Ivan-Velickovic (approved):

PR #219, opened 2025-02-27
Rebased version of:  Changes to enable linking with LLVM/lld #173 

kent-mcleod:
This PR is a rebased version of an older PR to enable linking with LLVM/lld: https://github.com/seL4/seL4_tools/pull/173/

The difference between the two is shown in the diff due to an earlier clean up of the riscv common linker file and changing the riscv stack allocation to take place in the linker file instead of coming from an object file:
```
diff --git a/elfloader-tool/src/linker.lds b/elfloader-tool/src/linker.lds
index c6d9338..20c06e8 100644
--- a/elfloader-tool/src/linker.lds
+++ b/elfloader-tool/src/linker.lds
@@ -4,13 +4,6 @@
  */
 
-/*
- * NOTE: For RISC-V 32-bit, having the .rodata section before the .data section causes the elfloader to
- *       freeze up. Thus, for 32-bit we move the .rodata section to after the .bss section as previously
- *       before some refactoring. The 64-bit version does not seem to be affect by this issue and uses thus
- *       uses the new layout.
- */
-
 #include <autoconf.h>
 #include <elfloader/gen_config.h>
 #include "image_start_addr.h"
@@ -29,7 +22,6 @@ SECTIONS
     {
         *(.text)
     }
-#ifdef CONFIG_ARCH_RISCV64
     . = ALIGN(16);
     .rodata :
     {
@@ -45,7 +37,6 @@ SECTIONS
         *(._archive_cpio)
         _archive_end = .;
     }
-#endif
     . = ALIGN(16);
     .data :
     {
@@ -57,34 +48,15 @@ SECTIONS
     . = ALIGN(16);
     .bss (NOLOAD) :
     {
-#ifdef CONFIG_ARCH_ARM
         . = ALIGN(0x1000);
         core_stack_alloc = .;
-        . += CONFIG_MAX_NUM_NODES * (1 << 12);
+        . = . + CONFIG_MAX_NUM_NODES * 1 << 12;
         core_stack_alloc_end = .;
-#endif
         _bss = .;
         *(.sbss*)
         *(.bss)
         *(.bss.*)
         _bss_end = .;
     }
-#ifdef CONFIG_ARCH_RISCV32
-    . = ALIGN(16);
-    .rodata :
-    {
-        *(.srodata*)
-        . = ALIGN(16);
-        *(.rodata)
-        *(.rodata.*)
-        /*
-         * ld crashes when we add this here: *(_driver_list)
-         */
-        . = ALIGN(16);
-        _archive_start = .;
-        *(._archive_cpio)
-        _archive_end = .;
-    }
-#endif
      _end = .;
 }
```


kent-mcleod:
This was tested with https://github.com/seL4/seL4/pull/1396 to show that all jobs pass successfully for sel4test and sel4bench across all platforms.

PR #222, opened 2025-03-05
elfloader,rpi4: Add smp-spin-table driver

kent-mcleod:
Very simple driver for booting multiple cores using the device tree tooling infrastructure.


lsf37 (approved):

lsf37:
(Style and gitlint are still failing)

PR #224, opened 2025-05-14
Arm: Reset CNTVOFF before switching to 32-bits

Indanz:
If no one else does, CNTVOFF will have a random value at boot, while MCS seL4 expects a low value.

This fixes the ZYNQMP106 CI failures with the zcu106 board.


Indanz:
Oh, come on:

```
# MISSING COPYRIGHT AND LICENSING INFORMATION

The following files have no licensing information:
* cmake-tool/LICENSE_BSD2.txt
* elfloader-tool/LICENSE_GPLv2.txt
```
They both have copyright info, it's just not at the very top...

Indanz:
Well, it looks like `.licenseignore` can be removed, as it seems to be ignored.

kent-mcleod:
It appears that the item is missing from the zynq32 overlay dts: https://github.com/seL4/seL4/blob/master/src/plat/zynqmp/overlay-zynqmp32.dts#L32 
It's present in the non-32 overlay.

lsf37:
> Well, it looks like `.licenseignore` can be removed, as it seems to be ignored.

Yes, that can be removed, as can be the two license text files it complain about -- they shouldn't be there in the first place (they should be in LICENSES/ which already does have a copy of both). The upgraded license checker no longer ignores license texts when they are not in the LICENSES directory, and it looks like we missed a few of those.

lsf37:
> It appears that the item is missing from the zynq32 overlay dts: https://github.com/seL4/seL4/blob/master/src/plat/zynqmp/overlay-zynqmp32.dts#L32 It's present in the non-32 overlay.

And that would explain why we are seeing this only on the 32-bit images.

Indanz:
> This is already reset for platforms that declare they have a generic timer in their device tree

Yes, for configs that have HYP enabled. Not for 32-bit running on 64-bit hardware.

My understanding is that when we run 32-bit with HYP disabled, Elfloaders executes elfloader-tool/src/arch-arm/32/crt0_64.S, which drops down into EL1 32-bit mode with EL2 disabled. At that point the generic timer code can't fix anything, as it doesn't have the rights.

I'll update the commit message to mention this.

> It appears that the item is missing from the zynq32 overlay dts

That should be added too, for 32-bit running on 32-bit hardware.  I don't know if we have any pure 32-bit HW with HYP support, that could be broken because of this missing.

Indanz:
This is needed for all platforms setting `ElfloaderArmV8LeaveAarch64`, if they don't have anything else zeroing `cntvoff`.

kent-mcleod:
> This is needed for all platforms setting `ElfloaderArmV8LeaveAarch64`, if they don't have anything else zeroing `cntvoff`. These are KernelPlatformImx8mm-evk, KernelPlatImx8mq, KernelPlatformImx8mp-evk and KernelPlatformRpi4.
> 
> Checking U-Boot code, it seems to only reset `cntvoff_el2` if it drops down from EL2 to EL1, like my patch does, or when dropping down from EL3 to EL2.

Ok, that makes sense to need to explicitly do it in the assembly that drops to EL1.

kent-mcleod (approved):

PR #230, opened 2025-06-05
Bump required cmake version

lsf37:
As decided in the most recent TSC meeting, bump the required cmake version to the latest available at the moment. Update command line options accordingly (`--max-subargs-per-line` option no longer exists).

This will change/complain about the style in many of our existing `cmake` files, but at some point we do need to switch, because to keep the old version working, we'd need to pin too many other packages to old versions and thereby downgrade other tools.

See also https://github.com/seL4/seL4/pull/1479

### DIFF
--- a/misc/style-cmake.sh
+++ b/misc/style-cmake.sh
@@ -14,7 +14,6 @@ PROGNAME=${0##*/}
 # to provide definitions for custom function formatting.
 CMAKE_FMT="--line-width 100 \
            --tab-size 4 \
-           --max-subargs-per-line 3 \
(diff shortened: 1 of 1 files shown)


Indanz (approved):
0.6.13 seems awfully specific, isn't it save to fix it to 0.6.x?

lsf37:
> Would be nice if the build system is stable for decades. Oh wait, that would be make...
> 
> (I still don't see how cmake is in any way better than make, other than being slightly faster, but let's not open that discussion here.)

I don't disagree, wasn't my first choice either :-)

PR #232, opened 2025-07-12
SBI fixes

Indanz:
With some luck it fixes the P550 boot problems at https://github.com/seL4/ci-actions/pull/420. I restrained myself to obvious improvements instead of any major changes.

The whole hard_id - logical core id stuff is extremely convoluted, both in the kernel as Elfloader, all to support a mapping that isn't 1:1. But the Elfloader code which calls `sbi_hart_start` assumes it is 1:1, so I'm not sure why we bother with all this. The only code needing FIRST_HART_ID is the (non-SMP?) plic code. Similarly, I don't understand why we go through all the complexity of switching harts if it isn't CONFIG_FIRST_HART, it doesn't matter which core does the shared init part. After synchronisation and jumping to the kernel it's all a big race anyway.

### DIFF
--- a/elfloader-tool/src/arch-riscv/boot.c
+++ b/elfloader-tool/src/arch-riscv/boot.c
@@ -152,7 +152,7 @@ static void set_and_wait_for_ready(int hart_id, int core_id)
 
     /* Wait untill all cores are go */
     for (int i = 0; i < CONFIG_MAX_NUM_NODES; i++) {
-        while (__atomic_load_n(&core_ready[i], __ATOMIC_RELAXED) == 0) ;
+        while (__atomic_load_n(&core_ready[i], __ATOMIC_ACQUIRE) == 0) ;
     }
 }
 #endif
--- a/elfloader-tool/src/arch-riscv/boot.c
+++ b/elfloader-tool/src/arch-riscv/boot.c
@@ -212,12 +212,11 @@ static int run_elfloader(UNUSED int hart_id, void *bootloader_dtb)
     /* Unleash secondary cores */
     __atomic_store_n(&secondary_go, 1, __ATOMIC_RELEASE);
 
-    /* Start all cores */
-    int i = 0;
-    while (i < CONFIG_MAX_NUM_NODES && hsm_exists) {
-        i++;
-        if (i != hart_id) {
-            sbi_hart_start(i, secondary_harts, i);
+    /* Start all other cores */
+    for (int i = 0; i < CONFIG_MAX_NUM_NODES && hsm_exists; ++i) {
+        int h = i + CONFIG_FIRST_HART_ID;
+        if (h != hart_id) {
+            sbi_hart_start(h, secondary_harts, h);
         }
     }
 
--- a/elfloader-tool/include/arch-riscv/sbi.h
+++ b/elfloader-tool/include/arch-riscv/sbi.h
@@ -24,8 +24,8 @@
     register uintptr_t a2 asm ("a2") = (uintptr_t)(arg2);   \
     register uintptr_t a7 asm ("a7") = (uintptr_t)(which);  \
     asm volatile ("ecall"                   \
-              : "+r" (a0)               \
-              : "r" (a1), "r" (a2), "r" (a7)        \
+              : "+r" (a0), "+r" (a1)        \
+              : "r" (a2), "r" (a7)          \
               : "memory");              \
     a0;                         \
 })
@@ -40,8 +40,8 @@
     register uintptr_t a6 asm ("a6") = (uintptr_t)(which);  \
     register uintptr_t a7 asm ("a7") = (uintptr_t)(extension); \
     asm volatile ("ecall"                   \
-              : "+r" (a0)               \
-              : "r" (a1), "r" (a2), "r" (a6), "r" (a7)      \
+              : "+r" (a0), "+r" (a1)        \
+              : "r" (a2), "r" (a6), "r" (a7)\
               : "memory");              \
     a0;                         \
 })


Indanz:
seL4's kernel `sbi_call` has the same problem actually, `a1` may be clobbered by the call, see https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/binary-encoding.adoc:

> All registers except a0 & a1 must be preserved across an SBI call by the callee.

Indanz:
Lovely, now HIFIVE_verification_SMP_gcc_64 fails with:
```
 Main entry hart_id:1
  Secondary entry hart_id:3 core_id:1
  Hart ID 3 core ID 1
  Secondary entry hart_id:2 core_id:2
  Hart ID 2 core ID 2
  Hart ID 1 core ID 0
  p_error: hart0: mepc=0x00000000800030b2 mstatus=0x0000000200001800
  sbi_trap_error: hart0: ra=0x00000000800079e2 sp=0x000000008001ff10
  sbi_trap_error: hart0: gp=0x0000000000000000 tp=0x0000000080020000
  sbi_trap_error: hart0: s0=0x000000008001ff20 s1=0x0000000000000000
  sbi_trap_error: hart0: a0=0x0000000080020060 a1=0x0000000000000002
  sbi_trap_error: hart0: a2=0x0000000000000003 a3=0x0000000000000000
  sbi_trap_error: hart0: a4=0x0000000000000001 a5=0x0000000080020060
  sbi_trap_error: hart0: a6=0x0000000000000001 a7=0x0000000000000000
  sbi_trap_error: hart0: s2=0x0000000080020000 s3=0x0000000000000000
  sbi_trap_error: hart0: s4=0x0000000080013008 s5=0x00000000800120e8
  sbi_trap_error: hart0: s6=0x0000000000000000 s7=0x0000000000000016
  sbi_trap_error: hart0: s8=0x0000000000002000 s9=0x0000000080013710
  sbi_trap_error: hart0: s10=0x0000000000000000 s11=0x0000000000000000
  sbi_trap_error: hart0: t0=0x0000000000000000 t1=0x0000000000000000
  sbi_trap_error: hart0: t2=0x0000000000000000 t3=0xffffffffffffffff
  sbi_trap_error: hart0: t4=0x000000000c200000 t5=0x0000000000000000
  sbi_trap_error: hart0: t6=0x0000000000000000
  [[Timeout]]
```
HIFIVE_verification_SMP_clang_64 also fails, but it doesn't cause a fault in the OpenSBI code. 

So I suspect that skipping hart ID 0 is needed for the HiFive, but that that will probably break polarfire, which is (perhaps because of that) not run for HW tests. Dropped that commit to see if the rest is okay.

axel-h:
> has a suspicious `Platform HART Count : 5`),

That number is fine. It has 4 application cores and one management core.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -215,6 +215,9 @@ static int run_elfloader(UNUSED int hart_id, void *bootloader_dtb)
     /* Start all cores */
```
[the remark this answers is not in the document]
Indanz: I think I'll change the code to initialise `i` to `CONFIG_FIRST_HART_ID`, as I think that's what makes most sense. And yes, expanding existing comment is better.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -212,10 +212,8 @@ static int run_elfloader(UNUSED int hart_id, void *bootloader_dtb)
     /* Unleash secondary cores */
     __atomic_store_n(&secondary_go, 1, __ATOMIC_RELEASE);
 
-    /* Start all cores */
-    int i = 0;
-    while (i < CONFIG_MAX_NUM_NODES && hsm_exists) {
-        i++;
+    /* Start all other cores */
+    for (int i = CONFIG_FIRST_HART_ID; i < CONFIG_MAX_NUM_NODES && hsm_exists; ++i) {
```
axel-h, line 216: The whole RISC-V boot seems a big legacy mess, specially this core lottery thing they came up with. I wonder if we should have a platform specific call here that has to make the decision to bring up a core or not. The default implementation would be to start all other cores except the current one. That would also remove the dependency on `CONFIG_FIRST_HART_ID` from the generic code here. Maybe we can even move this define then into a platform specific one, so we get rid of it as much as possible. Platforms should then provide some specific comment about their quirks.

Indanz: How did RISC-V manage to build up legacy mess, it's new! It had more than 50 years of computer science to look back at how not to do things, they have zero excuse for any mess they're causing with their bad choices.

The main thing missing is choosing which cores to run seL4 on. Now it's determined by CONFIG_FIRST_HART_ID and CONFIG_MAX_NUM_NODES, which seems reasonable to me.

How are we supposed to figure out the hart IDs of other cores and and which one we can use?  From what I understood, hart IDs are not guaranteed to be sequential? If it were, things are simple and we could also use `core ID == hart ID - CONFIG_FIRST_HART_ID`.

Indanz:
> seL4's kernel `sbi_call` has the same problem actually, `a1` may be clobbered by the call, see https://github.com/riscv-non-isa/riscv-sbi-doc/blob/master/src/binary-encoding.adoc:
> 
> > All registers except a0 & a1 must be preserved across an SBI call by the callee.

According to https://www.scs.stanford.edu/~zyedidia/docs/riscv/riscv-sbi.pdf, legacy extensions did guarantee that only `a0` would be touched, so it depends on whether it is a legacy extension call or a non-legacy extension call what the calling convention is. Legacy calls have EID 0x00-0x0F.

elfloader-tool/src/arch-riscv/boot.c
Indanz: Actually, if you go this way, it's better to have a platform specific `core_to_hart_id()` and `hart_to_core_id()` functions, with a default implementation that does hart ID == core ID.

HiFive can do core ID = hart ID - 1. No idea what Polarfire and Star64 need. Polarfire is an FPGA implementation, and Star64 is also based on a SiFive SoC, so probably the same as HiFive.

Ivan-Velickovic:
```
qemu-riscv-virt/config.cmake
14:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)

cheshire/config.cmake
14:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)

hifive/config.cmake
13:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 1)

spike/config.cmake
14:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)

polarfire/config.cmake
13:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 1)

star64/config.cmake
15:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 1)

ariane/config.cmake
13:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)

rocketchip/config.cmake
32:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)

eswin/config.cmake
14:    config_set(KernelPlatformFirstHartID FIRST_HART_ID 0)
```

To my understanding then Star64, Polarfire, and HiFive Unleashed need core ID = hart ID - 1, everything else is 'normal'.

> From what I understood, hart IDs are not guaranteed to be sequential?

I thought they were sequential, is that not the case?

Ivan-Velickovic:
> According to https://www.scs.stanford.edu/~zyedidia/docs/riscv/riscv-sbi.pdf, legacy extensions did guarantee that only a0 would be touched, so it depends on whether it is a legacy extension call or a non-legacy extension call what the calling convention is. Legacy calls have EID 0x00-0x0F.

All the kernel SBI calls are using the legacy calls so I don't think this is an issue?

elfloader-tool/src/arch-riscv/boot.c
Indanz: > The mhartid CSR is an MXLEN-bit read-only register containing the integer ID of the hardware thread running the code. This register must be readable in any implementation. **Hart IDs might not necessarily be numbered contiguously in a multiprocessor system**, but at least one hart must have a hart ID of zero. Hart IDs must be unique within the execution environment.

https://five-embeddev.com/riscv-priv-isa-manual/Priv-v1.12/machine.html#hart-id-register-mhartid

Only hartid 0 is guaranteed to be there, and they recommend to keep the hartids low numbers. However, the "RISC-V Supervisor Binary Interface Specification" specifies a mask for operations that operate on multiple harts, so they can't be too sparse.

Anything else other than sequentially numbering form 0 would be madness, but it's not guaranteed by the spec.

Indanz:
> Maybe just all the platforms we've dealt with so far happen to have sequential hart IDs...

Yes, and that will stay like that for the foreseeable future, until some weird hardware pops up and says "but the standard allows it..." and all existing software breaks. The standard should have the gut to say "No, unnecessary silly things aren't allowed". It's bad enough that you can't count on using hart ID 0 for S-mode.

Ivan-Velickovic (approved):

Indanz:
> If you've tested your patch I think it's fine to merge as is now.

Compiling RISCV with Clang fails for me, so I can't test if it fixes your original problem. Only tests done are the CI ones, I was hoping you could try it on the P550.

elfloader-tool/src/arch-riscv/boot.c
```diff
@@ -212,12 +212,11 @@ static int run_elfloader(UNUSED int hart_id, void *bootloader_dtb)
...
 
-    /* Start all cores */
-    int i = 0;
-    while (i < CONFIG_MAX_NUM_NODES && hsm_exists) {
-        i++;
-        if (i != hart_id) {
-            sbi_hart_start(i, secondary_harts, i);
+    /* Start all other cores */
+    for (int i = 0; i < CONFIG_MAX_NUM_NODES && hsm_exists; ++i) {
+        unsigned long h = i + CONFIG_FIRST_HART_ID;
```
axel-h, line 217:
```suggestion
        word_t h = i + CONFIG_FIRST_HART_ID;
```

Ivan-Velickovic:
@Indanz P550 is part of the sel4test CI already so ran on this PR when you added the hardware tests, Clang seems to pass https://github.com/seL4/seL4_tools/actions/runs/16296733479/job/46027209144.

Indanz:
> @Indanz P550 is part of the sel4test CI already so ran on this PR when you added the hardware tests, Clang seems to pass https://github.com/seL4/seL4_tools/actions/runs/16296733479/job/46027209144.

Sorry, I meant P550 SMP runs, the ones that got disabled by ttps://github.com/seL4/ci-actions/pull/420.

PR #235, opened 2025-08-01
simulate.py: don't unconditionally reset terminal

midnightveil:
The commit by @kent-mcleod f63e3d5b783431bdbfa5861dec6bc34311357f20 which added this says that "Qemu leaves the terminal in an inconsistent state" but personally I've never seen this, and whenever I run tests in simulate w/QEMU I need to always remember to remove this option from the script. I've seen coworkers also have to do this. The issue is that "tput reset" also clears the screen which means you can't see the log outputs.

Instead I've turned this into an option for --reset-terminal in case anyway actually wants to use this.


Indanz:
Yes, very annoying. I also always edit the simulate script to remove the sleep and the terminal reset.

PR #249, opened 2026-03-10
cmake style update

kent-mcleod:
- Don't dangle closing params onto new line
- 2 char indent instead of 4
- Drastically increase the number of elements that can be in an arg group before switching to vertical mode


Indanz (approved):
No opinion about 2 versus 4 tabs, I'd go whatever is used most in the current seL4 cmake files, but the rest sounds good.

lsf37 (approved):

PR #252, opened 2026-03-29
Workaround for what is likely a TX2 firmware bug

Indanz:
Depending on the size of the kernel, Elfloader causes SError traps. After a lot of testing and trying things out, it appears this is caused by speculative CPU memory accesses to device memory or something like that.

As this happens before Elfloader enables the MMU, this time it is not our mistake, but most likely a bug in the TX2 firmware or earlier bootloaders settings things up wrong.

The below change in no way fixes the underlaying bug or problem, but it does avoid triggering the fault.

### DIFF
--- a/elfloader-tool/src/linker.lds
+++ b/elfloader-tool/src/linker.lds
@@ -48,15 +48,17 @@ SECTIONS
     . = ALIGN(16);
     .bss (NOLOAD) :
     {
-        . = ALIGN(0x1000);
-        core_stack_alloc = .;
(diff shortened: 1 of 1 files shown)


Indanz:
Seems to break RISCV for some reason, I'm investigating.

Indanz:
Well, the work-around works for me locally, but then again, I didn't hit exactly the same problem that happens in CI, as I couldn't reproduce it with my build. I only managed to get the same kind of fault when adding some padding.

Oh well, was worth a try.

PR #253, opened 2026-03-30
Try mask serror and other interrupts

kent-mcleod:


lsf37 (approved):
Have tested that this fixes the TX2 problem. We should probably wait for the tests on the other boards to come back to make sure we haven't introduced anything new, but it looks good to me.

midnightveil:
I'd like to add a comment that we have code in this repo that *enables* SError for the TX2. https://github.com/seL4/seL4_tools/commit/c8cb3ebc9d9dbd6aadf9b6fe89a3c8065db58e33. I don't know why this was necesary.

lsf37:
> I'd like to add a comment that we have code in this repo that _enables_ SError for the TX2. [c8cb3eb](https://github.com/seL4/seL4_tools/commit/c8cb3ebc9d9dbd6aadf9b6fe89a3c8065db58e33). I don't know why this was necesary.

Classical commit message.. I don't know either, it might have been exactly to find cases such as the one we are working around now. We're not fixing the root cause, we're just adding a workaround for the more recent firmware apparently leaving something unmasked that was masked before.

Kent mentioned that the root cause is likely that the elf loader is not using the correct memory attributes in its mappings, which is something we should look at in the future.

lsf37:
There is a whole bunch of failures, but so far only boards refusing to boot or tests hitting a timeout (possibly also because of long boot time). TX2 does work. I think we can re-run the failed ones and merge this.

kent-mcleod:
> There is a whole bunch of failures, but so far only boards refusing to boot or tests hitting a timeout (possibly also because of long boot time). TX2 does work. I think we can re-run the failed ones and merge this.

Before merging we just need to update the comments that 0xf is DAIF bits all masked and confirm that having them all masked is intentional.

kent-mcleod:
Also I found that there was a different error altogether if I used a newer toolchain than the one in the docker container images.

lsf37:
> > There is a whole bunch of failures, but so far only boards refusing to boot or tests hitting a timeout (possibly also because of long boot time). TX2 does work. I think we can re-run the failed ones and merge this.
> 
> Before merging we just need to update the comments that 0xf is DAIF bits all masked 

That bit is easy

> and confirm that having them all masked is intentional.

What is the plan for that part?

The tests are ready now. If you sign off, it would otherwise be good to merge.

lsf37:
> Also I found that there was a different error altogether if I used a newer toolchain than the one in the docker container images.

I'd be fine with ignoring that for the release, but also something to look into. Do you think it might be the same Indan was seeing?

Indanz (approved):

Indanz:
> I'd like to add a comment that we have code in this repo that _enables_ SError for the TX2. [c8cb3eb](https://github.com/seL4/seL4_tools/commit/c8cb3ebc9d9dbd6aadf9b6fe89a3c8065db58e33). I don't know why this was necesary.

So it's all self-inflicted pain then, wonderful. I think the intention was to enable machine-check-exceptions caused by ECC, which would be delivered as SError. However, seL4 had special code to ignore SErrors for tx2 in the kernel, so what's the point of all this if you're going to ignore it anyway?

PR #254, opened 2026-04-03
fix elfloader: add UART MMIO mapping for debug output on AArch64 after MMU enable.

Whatever314:
Previously el1 error traps defined in <arch>/<mode>/debug.c doesn't print anything cos the uart memory regions are not mapped after enabling mmu. This is a preliminary fix for arm aarch64 platforms.

# Limitations
Sorry that I have neither time nor hardware to implement and test this fix on other platforms.


Indanz:
I'd rather disable the MMU in the trap code than this. That would make it possible to debug traps caused by bad MMU setups too.

PR #255, opened 2026-04-05
nanobp: tweaks for nanobp 0.4.9.x

lsf37:
nanobp 0.4.9.x declares its own nanobp target, so the one declared here is no longer needed (and conflicts with the one provided by nanobp).

Needs https://github.com/seL4/sel4test-manifest/pull/27 to work. The latter is a manifest PR, so "Test with" won't work for it.


lsf37:
(The test failure is expected)

Indanz (approved):
Cool, this means we can use the up-to-date version instead.

I tried fixing the compile error before, but my CMake fu wasn't great enough to figure out what's going on and what goes wrong where.

PR #256, opened 2026-05-02
elfloader: discard elf loader .eh_frame

lsf37:
- lld + clang on riscv32 generate a non-empty .eh_frame for libgcc.a with stack unwinding info for software-implemented div/mod. Other compiler + linker combinations only produce an empty .eh_frame (if at all).

  The non-empty .eh_frame ends up at the very beginning of the ELF file, displacing .start, and even though we are explicitly setting an entry point, OpenSBI is ignoring that entrypoint and just jumps to the first address, which no longer contains code. This means the boot hangs after OpenSBI.

  Placing the .eh_frame somewhere else explicitly fixes the problem, but we don't need it and discarding it is easier and more in line with what gcc produces. This affects the elfloader itself only, not the payload.

- disambiguate `x * 1 << 12` as `x * (1 << 12)`. I'm not sure if lld really interpreted this differently than GNU ld, but it is clearer in any case.

(Background of all of this is trying to get RISC-V clang builds in the new Trixie-based docker containers to work)

### DIFF
--- a/elfloader-tool/src/linker.lds
+++ b/elfloader-tool/src/linker.lds
@@ -50,7 +50,7 @@ SECTIONS
     {
         . = ALIGN(0x1000);
         core_stack_alloc = .;
-        . = . + CONFIG_MAX_NUM_NODES * 1 << 12;
+        . = . + CONFIG_MAX_NUM_NODES * (1 << 12);
         core_stack_alloc_end = .;
         _bss = .;
         *(.sbss*)
--- a/elfloader-tool/src/linker.lds
+++ b/elfloader-tool/src/linker.lds
@@ -59,4 +59,13 @@ SECTIONS
         _bss_end = .;
     }
      _end = .;
+
+    /* With lld on riscv32 a debug-only non-empty .eh_frame ends up ahead of
+      .start, and OpenSBI always jumps to the first address. We could place it
+      somewhere explicitly, but discarding it for elfloader is easier and safe
+      (application payload is unaffected). */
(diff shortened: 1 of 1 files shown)


lsf37:
The builds are failing because we need all of these for things to work

- https://github.com/seL4/seL4-CAmkES-L4v-dockerfiles/pull/96
- https://github.com/seL4/seL4_tools/pull/256
- https://github.com/seL4/seL4_tools/pull/257
- https://github.com/seL4/seL4_tools/pull/258

Indanz (approved):
I think the root problem is that the linker script may be too simple and doesn't specify what it actually wants very well, giving LDD room to do weird stuff that's unexpected, but technically allowed. If we used something like:

`.start IMAGE_START_ADDR :`

Or pinned the address of _start to IMAGE_START_ADDR, the code may still work, even with weird .eh_frames. That said, we still want to get rid of .eh_frame and other stuff like that. And if LDD places something before IMAGE_START_ADDR anyway, no idea what will happen then.

[This](https://maskray.me/blog/2020-12-19-lld-and-gnu-linker-incompatibilities) site has some incompatibilities between LD and LDD. One of them is that ld.lld places .rodata before .text. That may explain why .eh_frame ended up before .text.

lsf37:
I'm absolutely not opposed to improving the linker scripts, but I'm probably not the right person to do that.

Since we do want to skip the .eh_frame in any case and the tests are succeeding, I'm going to merge this one now, but happy to discuss further.

PR #259, opened 2026-05-03
make-uimage: better temp file handling

lsf37:
With recent clang builds, we can no longer assume that objcopy is GNU objcopy. It can also be llvm-objcopy, which does not work with /dev/stdout as output file.

Change the temp file handling so that we can directly give the temp file to objcopy without having to worry about race tmp conditions by creating a temp directory instead.

Test with: https://github.com/seL4/util_libs/pull/206

### DIFF
--- a/cmake-tool/helpers/make-uimage
+++ b/cmake-tool/helpers/make-uimage
@@ -58,9 +58,9 @@ written to OUTPUT-FILE.
 EOF
 }
 
-# Clean up temporary file.  $TEMPFILE is defined before this function is called.
+# Clean up temporary directory.  $TEMPDIR is defined before this function is called.
 cleanup () {
-    rm -f $TEMPFILE
+    rm -rf "$TEMPDIR"
 }
 
 # Output the start symbol from given ELF object.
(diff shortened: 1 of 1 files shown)


cmake-tool/helpers/make-uimage
```diff
@@ -58,9 +58,11 @@ written to OUTPUT-FILE.
 EOF
 }
 
-# Clean up temporary file.  $TEMPFILE is defined before this function is called.
+# Clean up temporary directory.  $TEMPDIR is defined before this function is called.
 cleanup () {
-    rm -f $TEMPFILE
+    if [ -n "${TEMPDIR:-}" ] && [ -d "$TEMPDIR" ]; then
```
Indanz: What is the purpose of the `:-` bit? It seems to replace the variable if it is empty, but there is nothing after -, so it will stay empty. Also, `rm -rf` doesn't error if the directory doesn't exist, so why check for that?

In other words, what's wrong with just `if [ -n "$TEMPDIR" ]; then`?

Indanz (approved):

cmake-tool/helpers/make-uimage
lsf37: `-n "$TEMPDIR"` will generate an error (edit:) when unset (`set -eu` is on), but `-n "${TEMPDIR:-}"` will not.

I guess I was a bit over-cautious there (rm -rf looks scary :-)). I wanted to catch empty and unset and the case where it is a file not a directory (and when it doesn't exist, although that is harmless as you say). Realistically, I don't actually see how any of these could happen, since the `trap` is only installed after TEMPDIR has been set. Maybe if someone would remove the directory while the script is running and the use the name as a file, but that would be extremely obscure.

If you think it's too much, we can also just drop this bit.

Indanz: I think it's too much and would make maintaining this script harder for no gain.

lsf37: Ok, I'll remove it

PR #260, opened 2026-05-05
tx2/platform_init: add missing smc memory clobber

lsf37:
The inline assembly in the send_smc function in tx2/platform_init.c only declares clobbered registers, but not memory. This leads more recent gcc versions to optimize too aggressively.

Add missing memory clobber (as we have in other smc/sbi calls) for the regs struct that carries the result of the smc call.

This fixes the current CI boot hang for TX2 release builds.


lsf37:
I've checked the other occurrences of `smc` in the sel4test scope (including kernel) to make sure we haven't missed any others. All others are fine. Same for `ecall` on RISCV for SBI calls.

midnightveil (approved):

Indanz (commented):
Glad you found this one this quickly, well done.

PR #261, opened 2026-05-07
opensbi: pass -std=gnu17 to opensbi build

lsf37:
GCC >= 15 defaults to c23 which forbids redefining bool, which is OpenSBI 0.9 does.

Until we can bump OpenSBI to >= 1.6 where this is fixed, pass -std=gnu17 (= gnu version of C18) to the OpenSBI build by defining a CC wrapper. The CC wrapper is working around missing way to pass in additional CFLAGS.

This is in preparation of updating the RISC-V gcc in our docker containers to 15.2.0. For earlier gcc versions (gcc >=8), this doesn't affect the build.

### DIFF
--- a/cmake-tool/helpers/rootserver.cmake
+++ b/cmake-tool/helpers/rootserver.cmake
@@ -119,11 +119,14 @@ function(DeclareRootserver rootservername)
           COMMAND make -s -C "${OPENSBI_PATH}" O="${OPENSBI_BINARY_DIR}"
                   PLATFORM="${KernelOpenSBIPlatform}" clean
           COMMAND ${CMAKE_OBJCOPY} -O binary "${elf_target_file}" "${OPENSBI_PLAYLOAD}"
+          # pass -std=gnu17 to the OpenSBI build for gcc >=15, can be dopped
+          # when OpenSBI is bumped to >= 1.6
           COMMAND
             make -C "${OPENSBI_PATH}" O="${OPENSBI_BINARY_DIR}"
-            CROSS_COMPILE=${CROSS_COMPILER_PREFIX} PLATFORM="${KernelOpenSBIPlatform}"
-            PLATFORM_RISCV_XLEN=${OPENSBI_PLAT_XLEN} PLATFORM_RISCV_ISA=${OPENSBI_PLAT_ISA}
-            PLATFORM_RISCV_ABI=${OPENSBI_PLAT_ABI} FW_PAYLOAD_PATH="${OPENSBI_PLAYLOAD}"
+            CROSS_COMPILE=${CROSS_COMPILER_PREFIX} "CC=${CROSS_COMPILER_PREFIX}gcc -std=gnu17"
+            PLATFORM="${KernelOpenSBIPlatform}" PLATFORM_RISCV_XLEN=${OPENSBI_PLAT_XLEN}
+            PLATFORM_RISCV_ISA=${OPENSBI_PLAT_ISA} PLATFORM_RISCV_ABI=${OPENSBI_PLAT_ABI}
+            FW_PAYLOAD_PATH="${OPENSBI_PLAYLOAD}"
           DEPENDS "${elf_target_file}" elfloader ${USES_TERMINAL_DEBUG})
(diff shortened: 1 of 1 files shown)


cmake-tool/helpers/rootserver.cmake
```diff
@@ -119,11 +119,14 @@ function(DeclareRootserver rootservername)
...
                   PLATFORM="${KernelOpenSBIPlatform}" clean
           COMMAND ${CMAKE_OBJCOPY} -O binary "${elf_target_file}" "${OPENSBI_PLAYLOAD}"
+          # pass -std=gnu17 to the OpenSBI build for gcc >=15, can be dopped
+          # when OpenSBI is bumped to >= 1.6
           COMMAND
             make -C "${OPENSBI_PATH}" O="${OPENSBI_BINARY_DIR}"
-            CROSS_COMPILE=${CROSS_COMPILER_PREFIX} PLATFORM="${KernelOpenSBIPlatform}"
-            PLATFORM_RISCV_XLEN=${OPENSBI_PLAT_XLEN} PLATFORM_RISCV_ISA=${OPENSBI_PLAT_ISA}
-            PLATFORM_RISCV_ABI=${OPENSBI_PLAT_ABI} FW_PAYLOAD_PATH="${OPENSBI_PLAYLOAD}"
+            CROSS_COMPILE=${CROSS_COMPILER_PREFIX} "CC=${CROSS_COMPILER_PREFIX}gcc -std=gnu17"
```
axel-h, line 126: The placement of the quote looks odd. Shouldn't this be:
```suggestion
            CROSS_COMPILE=${CROSS_COMPILER_PREFIX} CC="${CROSS_COMPILER_PREFIX}gcc -std=gnu17"
```

lsf37: It doesn't really matter where they go, but looking at the other ones, they do only put them on the rhs, so I'm happy to change it for consistency.

lsf37: only that this actually breaks the build. I'll have a closer look later.

Indanz (approved):
I don't like that we need knowledge about the compiler and OpenSBI version in seL4_tools, but so be it.

Would be nice if this was behind a compiler and OpenSBI version check though, so things don't break when we do update OpenSBI and we forget to remove this again. But adding this is probably more hassle than fixing future breakage I guess.

After Axel's suggestion is applied, it seems ready to go.

Better solution would be to update OpenSBI, what are the problems with that?

lsf37:
> I don't like that we need knowledge about the compiler and OpenSBI version in seL4_tools, but so be it.
> 
> Would be nice if this was behind a compiler and OpenSBI version check though, so things don't break when we do update OpenSBI and we forget to remove this again. 

It should keep working actually. What OpenSBI did is an ifdef guard for that bool declaration that fails. So the danger is more that we forget about it forever, that's why I put in the comment from which OpenSBI version on we can drop this again.

> Better solution would be to update OpenSBI, what are the problems with that?

We need the RFC for SBI calls in seL4 first, then we can upgrade (more recent OpenSBI hide the timer register).

(Edit: this could be fairly soon if we implement the SBI RFC soon)

cmake-tool/helpers/rootserver.cmake
lsf37: I might have gotten this correct by accident initially -- it does matter where the quotes are. My version happens to be a quoted argument (https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#quoted-argument), and when the quotes are on the right, it is an unquoted argument with legacy support for quotes, which apparently is not the same.

`CC="${CROSS_COMPILER_PREFIX}gcc -std=gnu17"` produces `CC=${CROSS_COMPILER_PREFIX}gcc\ -std=gnu17`, which explains the error message `riscv64-unknown-elf-gcc -std=gnu17: not found` in that version.

I suspect that if any of the other ones expand to something with space in them, they would also break.

lsf37: (I've reverted the change now)

PR #264, opened 2026-05-20
elfloader(aarch64): fix MemAttr for stage 2 PTs

midnightveil:
Please see the adjacent seL4 PR to correct the same behaviour in seL4, as well as the one for the Microkit regarding the choices of these values.

**NOTE**: This currently breaks seL4 booting.

Test with: https://github.com/seL4/seL4/pull/1668


midnightveil:
Please see my comment here: https://github.com/seL4/seL4/pull/1668#issuecomment-4496238154.

tldr; Stage 1 translation is used for EL2, not Stage 2.

PR #265, opened 2026-07-03
helpers: use posix invocation for `find`

lsf37:
`find ... -printf "%p "` is a GNU extension. `-exec printf "%s " {}` does the same thing and works on BSD systems as well.

(One more step in the quest to natively build on Mac)


Indanz (approved):

Indanz:
I did test the command manually to see if the result is the same, and it is.

PR #266, opened 2026-07-27
application_settings: add UserOptimisation

### LINKED ISSUE #1436: clang rv64/rv32 simulations started taking 12min+
Something I noticed recently was that clang's rv64imac and rv32imac simulation seems to have started taking 12min+, which is much longer than most of the other simulation runs (4-5 mins).

This appears to have been since #1392 based off the main branch's CI times, though it could have been some other coincident change (not that I can see anything relevant).

|          | Times | Link |
|----------|-------|------|
| After    | 4min for gcc-rv64imac, 12min for clang-rv64imac | https://github.com/seL4/seL4/actions/runs/13452246363/job/37689305232 |
| Before    | 4min for gcc-rv64imac, 6min for clang-rv64imac | https://github.com/seL4/seL4/actions/runs/13380420037/job/37574500124 |

- [ ] Why longer?
- [x] ~~Also, I think these should be rv64imadfc now?~~ https://github.com/seL4/ci-actions/pull/385

lsf37:
Add `UserOptimisation` setting that determines user space compiler optimisation level for CMake RELEASE mode and `ApplyUserOptimisationSettings ()` function to apply the setting.

Default CMake RELEASE mode settings are `-O3 -DNDEBUG`, so the function also sets `-DNDEBUG`.

Default to -O2, because -O3 tends to be too aggressive.

Apply by default in `ApplyCommonReleaseVerificationSettings()` which is used in all seL4 tooling projects.

This might fix https://github.com/seL4/seL4/issues/1436

### DIFF
--- a/cmake-tool/helpers/application_settings.cmake
+++ b/cmake-tool/helpers/application_settings.cmake
@@ -95,7 +95,19 @@ function(ApplyCommonSimulationSettings kernel_sel4_arch)
   endif()
 endfunction()
 
+# Set the optimisation level for user space for cmake RELEASE mode.
+# (The default cmake RELEASE flags otherwise are `-O3 -NDEBUG`).
+function(ApplyUserOptimisationSettings)
+  set(UserOptimisation "-O2" CACHE STRING "Set the user mode optimisation level")
+  set_property(CACHE UserOptimisation PROPERTY STRINGS "-O2;-Os;-O0;-O1;-O3")
+  set(CMAKE_C_FLAGS_RELEASE "${UserOptimisation} -DNDEBUG" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_FLAGS_RELEASE "${UserOptimisation} -DNDEBUG" CACHE STRING "" FORCE)
+  set(CMAKE_ASM_FLAGS_RELEASE "${UserOptimisation} -DNDEBUG" CACHE STRING "" FORCE)
+endfunction()
+
 function(ApplyCommonReleaseVerificationSettings release verification)
+  ApplyUserOptimisationSettings()
+
   # Setup flags for different combinations of 'release' (performance optimized builds) and
   # 'verification' (verification friendly features) builds
   if(release)


lsf37:
If I understand cmake correctly, you should be able to call `-DUserOptimisation=-O1` from the outside now.

midnightveil:
It doesn't look like this has made any difference to the simulation run time.

lsf37:
> It doesn't look like this has made any difference to the simulation run time.

Yes, no luck it seems. We could still add this for other reasons, but it does not fix the clang simulation time.

Indanz (commented):
CMake already has multiple ways of specifying compiler flags and users know about them already (like `CMAKE_C_FLAGS`). Adding a new seL4 specific option is not the way to go for this I think. In this case it also only changes the release settings, which is not obvious from the setting's name.

I think the users of the projects should set the correct flags during configure. In our case, that would be CI.

If we want we can change the default flags to something else, but we should only do that if those flags weren't specified by the user already.

CMake seems a bit of a mess, I don't know what the modern recommended way of handling this is.

The [Arch Linux wiki](https://wiki.archlinux.org/title/CMake_package_guidelines#CMake_can_automatically_override_the_default_compiler_optimization_flag) point out that specifying the build type will add flags like `CMAKE_C_FLAGS_RELEASE` which will overrule the user's CFLAGS because they are appended last. I think we set it to none for the kernel already somewhere.

[This site](https://alexreinking.com/blog/how-to-use-cmake-without-the-agonizing-pain-part-2.html) basically says we do everything horribly wrong, but other than that also says to use presets for things like this instead of messing with the flags directly.

lsf37:
>  In this case it also only changes the release settings, which is not obvious from the setting's name.

Yes, good point, if we want to keep it at all we should definitely rename it.

(Edit: or maybe extend the setting to debug mode as well as we do for the kernel. That'd probably be better)

Ok, so overall this doesn't fix what we wanted and it doesn't seem to be the right way to go. Should we just close it? (Happy to, this was mostly an experiment).

For what the Arch Linux wiki says: yes, `CMAKE_C_FLAGS_RELEASE` is exactly what I'm tweaking (+ CXX + ASM, because we use those as well). We could set them to none and add our own, but that doesn't really change the result in the end.

I don't know of any other way to change the default, though (than either this or equivalently setting to none and adding our own option), precisely because CMake will force its own `CMAKE_C_FLAGS_RELEASE` as soon as you set release mode. So you do have to do something to it.

midnightveil:
Yeah, I think we can probably just close it.

Indanz:
> For what the Arch Linux wiki says: yes, `CMAKE_C_FLAGS_RELEASE` is exactly what I'm tweaking (+ CXX + ASM, because we use those as well). We could set them to none and add our own, but that doesn't really change the result in the end.

It would avoid the awkward interaction between `CMAKE_C_FLAGS_RELEASE` and user's `CMAKE_C_FLAGS`, with the former overriding the latter.

> I don't know of any other way to change the default, though (than either this or equivalently setting to none and adding our own option), precisely because CMake will force its own `CMAKE_C_FLAGS_RELEASE` as soon as you set release mode. So you do have to do something to it.

The default is only applicable if you do set `CMAKE_BUILD_TYPE` to something.

I think setting `CMAKE_C_FLAGS_RELEASE` or `CMAKE_C_FLAGS` to what we want is fine, but we should first check whether the user set it and leave it unchanged if it's already set. That is, there is no need for a separate option for this, as we can just check the existing variables already.

The alternative to the release/debug build type would be presets like mentioned in that link, but I don't know how that would improve things compared to the above.

> Yes, no luck it seems. We could still add this for other reasons, but it does not fix the clang simulation time.

Have you verified that the compile options are as expected? (With a clean build.)

lsf37:
> > I don't know of any other way to change the default, though (than either this or equivalently setting to none and adding our own option), precisely because CMake will force its own `CMAKE_C_FLAGS_RELEASE` as soon as you set release mode. So you do have to do something to it.
> 
> The default is only applicable if you do set `CMAKE_BUILD_TYPE` to something.
> 
> I think setting `CMAKE_C_FLAGS_RELEASE` or `CMAKE_C_FLAGS` to what we want is fine, but we should first check whether the user set it and leave it unchanged if it's already set. That is, there is no need for a separate option for this, as we can just check the existing variables already.

I can have a look if/how that works. There's a lot of reading/learning needed every time before I do something with CMake, but at some point I hope it will get better..


> The alternative to the release/debug build type would be presets like mentioned in that link, but I don't know how that would improve things compared to the above.
> 
> > Yes, no luck it seems. We could still add this for other reasons, but it does not fix the clang simulation time.
> 
> Have you verified that the compile options are as expected? (With a clean build.)

Yes, they were all -O2 (both on clean build and rebuild)

PR #267, opened 2026-09-16
EFI: Add missing _end for AArch64

Indanz:
Hopefully fixes the boot problem we have for ROCKPRO64.


lsf37 (approved):

midnightveil (approved):
Looks like it works now.

Indanz:
Weird that a missing symbol would resolve to -1 instead of giving a linker error though.

PR #268, opened 2026-09-18
cmake-tool: add ARM_PLAT aliases to elfloader binary_list

ShankaradityaNS:
`ApplyData61ElfLoaderSettings()` takes a platform name and checks it against `binary_list`, `efi_list` and `uimage_list`. Some projects call it with `${KernelARMPlatform}` rather than `${KernelPlatform}` — `camkes-vm-examples/settings.cmake:73`, for instance.

For most platforms those two variables are identical, but the kernel sets them differently for four boards:

| KernelPlatform | KernelARMPlatform |
|---|---|
| bcm2837 | rpi3 |
| bcm2711 | rpi4 |
| bcm2712 | rpi5 |
| rk3568 | rock3b |

`binary_list` contains only the `KernelPlatform` names, so when called with the ARM_PLAT name these four fall through to `ElfloaderImage=elf`. U-Boot's `go` then executes the ELF header and aborts.

This adds the ARM_PLAT aliases next to their existing twins. Purely additive — nothing removed or reordered, and no other platform's behaviour changes.

Verified on a Raspberry Pi 4B: before, `ElfloaderImage:STRING=elf` and the board aborted on `7f45 4c46`; after, `binary`, and it boots. I don't have rpi3, rpi5 or rock3b hardware, so those are fixed by the same reasoning rather than by test.


midnightveil:
Was this PR written by AI?

Indanz:
> Some projects call it with `${KernelARMPlatform}` rather than `${KernelPlatform}`

This is the real bug, please fix that instead of working around it here.

